PR #9, opened 2015-06-08
1.5 review

ecino:
# 1.5 Improvements

Contains new features added after split of contracts.

## Sponsorships / Contracts

- Dispatch GP Sync functionalities between modules sponsorship_sync and
  child_sync.
- New module sponsorship_switzerland with tailors the sponsorships for
  specificities of our country. It does not contain any synchronization with
  GP.
- Adapt to changes of contracts made in repository compassion-modules and
  compassion-accounting !

## Others

- Copy bvr_reference of move_lines into invoices when generated at validation
- Removed module partner_compassion and replaced by partner_switzerland with
  only switzerland specific code


CSester:
Should I look at this as Travis build fails ?

ecino:
Yes please, it's just wrong headers in a small module, the fix will be quick.

child_sync_typo3/wizard/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+#    @author: Emanuel Cino <ecino@compassion.ch>
+#
+#    The licence is in the file __openerp__.py
+#
+##############################################################################
+
+from . import child_on_typo3_wizard
+from . import child_remove_from_typo3
+from . import delegate_child_wizard
+from . import child_depart_wizard
```
CSester: Missing empty line

CSester:
:+1: 

My only concern is about the fast growing hardcoded string comparison usage rather than constants usage (and this in all repos). What is the guideline ?

ecino:
You are right this is not good. I think David removed a lot of those strings while moving the code (and adding fields like category_name for products). I will go through these constants and try to rewrite them properly. Thanks for the review !

ecino:
@CSester I removed the constants for all product gift types, and also for bvr_adherent_number (in compassion-modules). I also fixed the travis build for this one. Please tell me if you find other constants that should not be hardcoded.

PR #15, opened 2015-08-28
Fix

ndtran:
fix bug display child when run wizard


ecino:
Thanks! I wait for the fix of typo3 before merging.

PR #17, opened 2015-11-11
add import_spoken_lang module

rz-comp:
Import spoken language data from GP.
Data is imported in two groups:
- For every partner, the set of languages that the person speaks
- For every country, the set of languages that are spoken there

### DIFF
--- a/import_spoken_lang/__init__.py
+++ b/import_spoken_lang/__init__.py
@@ -1,0 +1,11 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2015 Compassion CH (http://www.compassion.ch)
+#    Releasing children from poverty in Jesus' name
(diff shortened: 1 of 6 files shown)


import_spoken_lang/__init__.py
```diff
@@ -0,0 +1,11 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2015 Compassion CH (http://www.compassion.ch)
+#    Releasing children from poverty in Jesus' name
+#    @author:
```
ecino: No author ?

ecino:
Please add README file to add module description. Follow the same structure as in other modules.

ecino:
Can you add in the logs a progress info ?

ecino:
:+1:

PR #18, opened 2016-01-13
Update partner views for sbc_compassion

ecino:
- update sponsorship preferred languages after importing spoken languages


rz-comp:
:+1:

PR #20, opened 2016-01-29
Rename sendgrid.email -> mail.mail

rz-comp:


ecino:
Can you please also change in `contracts.py` and `partner_compassion.py`? I added sending e-mails from there.

PR #35, opened 2016-04-06
Import letters from NAS

MickSandoz:
add option for import letters from a shared location on a NAS
imported letters are stored on NAS in a folder specify in data file scan_letter_params.xml

### DIFF
--- a/multi_attachment_switzerland/__openerp__.py
+++ b/multi_attachment_switzerland/__openerp__.py
@@ -29,10 +29,6 @@
 {
     'name': "multi_attachment_switzerland",
     'summary': """ Upload multi-attachment""",
-    'description': """ Upload multi-attachment:
-         - specify in qweb.xml models where multiple selection is possible
-         - sync_mail_multi_attach module must be in addons on server
-           https://www.odoo.com/apps/modules/8.0/sync_mail_multi_attach/""",
     'author': "Michael Sandoz",
     'website': "http://www.compassion.ch",
 
--- a/multi_attachment_switzerland/__openerp__.py
+++ b/multi_attachment_switzerland/__openerp__.py
@@ -1,0 +1,50 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#       ______ Releasing children from poverty      _
+#      / ____/___  ____ ___  ____  ____ ___________(_)___  ____
+#     / /   / __ \/ __ `__ \/ __ \/ __ `/ ___/ ___/ / __ \/ __ \
+#    / /___/ /_/ / / / / / / /_/ / /_/ (__  |__  ) / /_/ / / / /
+#    \____/\____/_/ /_/ /_/ .___/\__,_/____/____/_/\____/_/ /_/
+#                        /_/
+#                            in Jesus' name
+#
+#    Copyright (C) 2014 Compassion CH (http://www.compassion.ch)
+#    @author: Michael Sandoz <sandozmichael@hotmail.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': "multi_attachment_switzerland",
+    'summary': """ Upload multi-attachment""",
+    'description': """ Upload multi-attachment:
+         - specify in qweb.xml models where multiple selection is possible
+         - sync_mail_multi_attach module must be in addons on server
+           https://www.odoo.com/apps/modules/8.0/sync_mail_multi_attach/""",
+    'author': "Michael Sandoz",
+    'website': "http://www.compassion.ch",
+
+    # Categories can be used to filter modules in modules listing
+    # Check https://github.com/odoo/odoo/blob/master/openerp/addons/base/
+    # module/module_data.xml
+    # for the full list
+    'category': 'Social Network',
+    'version': '0.1',
+
+    # WARNING sync_mail_multi_attach module necessary for this one
+    # to work correctly
+    'depends': ['sync_mail_multi_attach'],
+    'qweb': ['static/src/xml/qweb.xml'],
+}
--- a/sbc_email/__openerp__.py
+++ b/sbc_email/__openerp__.py
@@ -36,6 +36,7 @@
     'data': [
         'data/email_text_templates.xml',
         'data/scan_letter_params.xml',
+        'views/import_config_view.xml',
         'views/import_letters_history_view.xml',
         'views/sbc_email_view.xml',
     ],
--- a/sbc_email/__openerp__.py
+++ b/sbc_email/__openerp__.py
@@ -35,6 +35,7 @@
     'depends': ['sbc_compassion', 'mail_sendgrid'],
     'data': [
         'data/email_text_templates.xml',
+        'data/scan_letter_params.xml',
         'views/sbc_email_view.xml',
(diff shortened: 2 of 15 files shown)


sbc_email/models/import_letters_history.py
```diff
@@ -0,0 +1,438 @@
...
+    ##########################################################################
+    @api.multi
+    @api.onchange(
+        "data", "import_line_ids", "letters_ids", "from_imports_nas_folder")
+    def _count_nber_letters(self):
+        """
+        Counts the number of scans. If a zip file is given, the number of
+        scans inside is counted.
+        """
+        for inst in self:
```
ecino: onchange methods are usually called on a single record. You can therefore start your method with `self.ensure_one()` and remove the iteration over `self`, and use `self` instead of `inst`.

sbc_email/models/import_letters_history.py
```diff
@@ -0,0 +1,438 @@
...
+            if inst.state in ("open", "pending", "ready"):
```
ecino: You can simplify your code by doing:
```
if self.state and self.state != 'draft':
    super(ImportLettersHistory, self)._count_nber_letters()
else:
    ... (your code from line 65)
```

sbc_email/models/import_letters_history.py
```diff
@@ -0,0 +1,438 @@
...
+                inst.nber_letters = len(inst.import_line_ids)
+            elif inst.state == "done":
+                inst.nber_letters = len(inst.letters_ids)
+            elif inst.state is False or inst.state == "draft":
+                # counter
+                tmp = 0
+                # if files are selected by user
+                if not inst.from_imports_nas_folder:
+                    # loop over all the attachments
```
ecino: You can remove all the code inside this condition and replace by calling super method.

sbc_email/models/import_letters_history.py
```diff
@@ -0,0 +1,438 @@
...
+                        for sharedFile in listPaths:
+                            if func.check_file(sharedFile.filename) == 1:
+                                tmp += 1
+                            elif func.isZIP(sharedFile.filename):
+                                logger.info(
+                                    'File to retrieve: {}'.format(
+                                        imported_letter_path +
+                                        sharedFile.filename))
+
+                                with tempfile.NamedTemporaryFile(
```
ecino: You don't need a temporary file, you can directly load the smb file inside a buffer

sbc_email/models/import_letters_history.py
```diff
@@ -0,0 +1,438 @@
...
+            if self.from_imports_nas_folder:
+                letters_import.state = 'pending'
+                if self.env.context.get('async_mode', True):
+                    session = ConnectorSession.from_env(self.env)
+                    ilh.import_letters_job.delay(
+                        session, self._name, letters_import.id)
+                else:
+                    letters_import._run_analyze()
+            # End code Michael Sandoz
+            elif letters_import.data:
```
ecino: You can do `else: super(...)`

ecino:
I won't comment at each method but you should never copy-paste a method from its parent class! Use super() method instead to call it in the situation when default behaviour is the same and do only the modifications needed in your inherited method.

If you copy-paste, the code won't be maintainable (we will have to change at different locations the same code).

sbc_email/data/scan_letter_params.xml
```diff
@@ -0,0 +1,18 @@
+<openerp>
+	<data noupdate="1">
+		<record id="share_on_nas" model="ir.config_parameter">
+			<field name="key">sbc_email.share_on_nas</field>
+			<field name="value">Public</field>
+		</record>
+
+		<record id="scan_letter_imports" model="ir.config_parameter">
```
ecino: You can remove this record.

sbc_email/models/import_letters_history.py
```diff
@@ -0,0 +1,384 @@
...
+            smb_user = config.get('smb_user')
+            smb_pass = config.get('smb_pwd')
+            smb_ip = config.get('smb_ip')
+            smb_port = int(config.get('smb_port', 0))
+            if not (smb_user and smb_pass and smb_ip and smb_port):
+                return False
+
+            smb_conn = SMBConnection(smb_user, smb_pass, 'openerp', 'nas')
+            config_obj = self.env['ir.config_parameter']
+            share_nas = (config_obj.search(
```
ecino: you can use `self.env.ref('sbc_email.share_on_nas')`

sbc_email/models/import_letters_history.py
```diff
@@ -0,0 +1,384 @@
...
+            [('key', '=', 'sbc_email.share_on_nas')])[0]).value
+
+        # Retrieve SMB configuration
+        smb_user = config.get('smb_user')
+        smb_pass = config.get('smb_pwd')
+        smb_ip = config.get('smb_ip')
+        smb_port = int(config.get('smb_port', 0))
+        if not (smb_user and smb_pass and smb_ip and smb_port):
+            return False
+        smb_conn = SMBConnection(smb_user, smb_pass, 'openerp', 'nas')
```
ecino: You could put your connection code inside a method since you use it twice in the class (to avoid duplicate code).

ecino:
Can you update the README file of sbc_email module in order to add the description of the system parameters you added?

You can also edit the module description and title to better match what it is doing now.

ecino:
We talked about creating xml data for the config. Could you add 4 records for 4 import configs taking in 4 different folders ?
The 4 configs should be:

1. Standard
2. Standard with attachments (field attachments must be set to "send by mail")
3. Non-Standard (template must be set to "Default Template")
4. Non-Standard with attachments (field attachments must be set to "send by mail" and template must be set to "Default Template")

sbc_email/data/scan_letter_params.xml
MickSandoz: It's used to save manul import.
Would you prefer that I use a "config import" just for that case or keep it in config?

sbc_email/models/import_letters_history.py
MickSandoz: I tried to do it but in method where self.ensure_one() is called, it's doesn't work...
I will try again

MickSandoz:
OK for the 2 last global comments
I will do it

sbc_email/models/import_letters_history.py
```diff
@@ -0,0 +1,384 @@
...
+    #                             FIELDS METHODS                             #
+    ##########################################################################
+    @api.one
+    @api.onchange("data", "import_folder_path")
+    def _count_nber_letters(self):
+        """
+        Counts the number of scans. If a zip file is given, the number of
+        scans inside is counted.
+        """
+        self.ensure_one()
```
ecino: `ensure_one()` is useless if you put `api.one` as decorator.

sbc_email/data/scan_letter_params.xml
ecino: Ok :+1:

multi_attachment_switzerland/__openerp__.py
```diff
@@ -0,0 +1,50 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': "multi_attachment_switzerland",
+    'summary': """ Upload multi-attachment""",
+    'description': """ Upload multi-attachment:
```
ecino: Description is the old way for doing module description. You should include a README and Odoo will parse it to display its content in the module page.

sbc_email/__openerp__.py
```diff
@@ -28,13 +28,19 @@
 ##############################################################################
 {
     'name': 'Sponsor to beneficiary email communication',
+    'description': """ Use to manage Sponsor to beneficiary email communication
```
ecino: You should not add a description in the file, the README is used for that.

sbc_email/data/import_config_templates.xml
```diff
@@ -0,0 +1,24 @@
+<openerp>
+    <data noupdate="1">
+        <record id="standard" model="import.letter.config">
+            <field name="name">Standard</field>
+            <field name="import_folder_path">\Scan Lettres S2B\Sandard</field>
+        </record>
+        <record id="standard_with_attachments" model="import.letter.config">
+            <field name="name">Standard avec pièces-jointes</field>
```
ecino: You should put names in English

PR #37, opened 2016-04-07
Send correpondence on local translate platform if needed.

eicher31:
And we recover it after translation.
We have updateted mysql_connector for more generic.


sbc_translation/models/correspondence.py
```diff
@@ -22,22 +33,119 @@ class Correspondence(models.Model):
...
+        """ Create a message for sending the CommKit after be translated on
+             the local translate plaforme.
+        """
         sponsorship = self.env['recurring.contract'].browse(
             vals['sponsorship_id'])
-        if sponsorship.reading_language.id == vals['original_language_id']:
-            letter = super(Correspondence, self).create(vals)
-        else:
+
+        original_lang_id = self.env['res.lang.compassion'].browse(
```
ecino: You could start the method with `letter = super(..)` so that you can then user `original_lang = letter.original_language_id` and `sponsorship = letter.sponsorship_id`.

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +33,119 @@ class Correspondence(models.Model):
...
+            vals['original_language_id'])
+
+        if original_lang_id.translatable:
```
ecino: I think this test is wrong. If we can translate a language does not mean that we have to! For instance if the letter is in french, `translatable` will be `True`, but if the letter is for Togo, we don't need to translate it.

sbc_translation/models/correspondence.py
ecino: Forget it, I didn't see you don't use the same constructor when letter has to be translated...

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +33,119 @@ class Correspondence(models.Model):
...
             letter = super(Correspondence, self.with_context(
                 no_comm_kit=True)).create(vals)
-            self.alternative()
+            self.send_local_translate(letter, sponsorship)
```
ecino: Why not doing `letter.send_local_translate()` and remove letter and sponsorship from the method parameters?

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +33,119 @@ class Correspondence(models.Model):
...
+                    self.env['gmc.message.pool'].create({
+                        'action_id': action_id,
+                        'object_id': letter["letter_odoo_id"]
+                    })
+
+                tc.remove_letter(letter["id"])
+
+    def get_child_langs_code(self, sponsorship):
+        """ Return children's code iso """
+        codes = []
```
ecino: I think you don't need a method for that as you can replace by:
`sponsorship.child_id.project_id.country_id.spoken_lang_ids.mapped('code_iso')`

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +33,119 @@ class Correspondence(models.Model):
...
+    @api.model
+    def check_local_translation_done(self):
+        tc = translate_connector.TranslateConnect()
+        letters_to_update = tc.get_translated_letters()
+
+        if not letters_to_update == -1:
+            for letter in letters_to_update:
+                correspondence = self.browse(letter["letter_odoo_id"])
+                reload(sys)
+                sys.setdefaultencoding('UTF8')
```
ecino: You should do this at the beginning of the method to avoid reloading at each iteration.

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +33,119 @@ class Correspondence(models.Model):
...
+                tg_lang = letter["target_lang"]
+                target_lang_id = self.env['res.lang.compassion'].search(
+                    [('code_iso', '=', tg_lang)]).id
+                # find the good text to writte
+                if tg_lang == 'eng':
+                    target_text = 'english_text'
+                elif tg_lang in self.get_child_langs_code(correspondence):
+                    target_text = 'translated_text'
+                else:
+                    logger.error("Correspondence letters language no match")
```
ecino: Would a `raise` be more approriate? It would fire the error. Or if this should not be blocking you could add a `continue` statement to go to the next letter?

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +33,119 @@ class Correspondence(models.Model):
...
+
+                # UPDATE Odoo Database
+                correspondence.write(
+                    {target_text: letter["text"],
+                     'state': 'Received in the system',
```
ecino: Will the support of child letters be added in a next Pull request?
Because this won't work for child letters, they should be in `Published to Global Partner` state.

sbc_translation/models/translate_connector.py
```diff
@@ -0,0 +1,157 @@
...
+    """ Contains all the utility methods needed to talk with the MySQL server
+        used by translate platform, as well as all mappings
+        from OpenERP fields to corresponding MySQL fields.
+    """
+
+    def __init__(self):
+        super(TranslateConnect, self).__init__('mysql_translate_host',
+                                               'mysql_translate_user',
+                                               'mysql_translate_pw',
+                                               'mysql_translate_db')
```
ecino: You could avoid all these lines by formatting differently.

sbc_translation/models/correspondence.py
eicher31: Yes

sbc_translation/models/correspondence.py
```diff
@@ -8,8 +8,20 @@
 #    The licence is in the file __openerp__.py
 #
 ##############################################################################
+import sys
+import base64
+import pdb
```
ecino: No pdb please

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +34,114 @@ class Correspondence(models.Model):
...
             vals['sponsorship_id'])
-        if sponsorship.reading_language.id == vals['original_language_id']:
-            letter = super(Correspondence, self).create(vals)
-        else:
+
+        original_lang = self.env['res.lang.compassion'].browse(
+            vals['original_language_id'])
+
+        if original_lang.translatable and original_lang.id not in sponsorship\
+                .child_id.project_id.country_id.spoken_lang_ids.mapped('id'):
```
ecino: You can simply test `... and original_lang not in sponsorship.child_id.project_id.country_id.spoken_lang_ids`

sbc_translation/data/nas_parameters.xml
```diff
@@ -0,0 +1,13 @@
+<openerp>
+  <data noupdate="1">
+    <record id="sbc_translation.nas_share_name" model="ir.config_parameter">
+      <field name="key">sbc_translation.nas_share_name</field>
+      <field name="value">GP</field>
+    </record>
+    
+    <record id="sbc_translation.nas_letters_store_path" model="ir.config_parameter">
```
ecino: You don't need to put module name in xml record_id. `nas_letters_store_path` is enough.

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +34,114 @@ class Correspondence(models.Model):
...
+        if not (smb_user and smb_pass and smb_ip and smb_port):
+            raise Exception('No config SMB in file .conf')
+
+        # Copy file in the imported letter folder
+        smb_conn = SMBConnection(smb_user, smb_pass, 'openerp', 'nas')
+        if smb_conn.connect(smb_ip, smb_port):
+            file_ = BytesIO(base64.b64decode(
+                self.letter_image.with_context(
+                    bin_size=False).datas))
+            config_obj = self.env['ir.config_parameter']
```
ecino: When you define datas in xml, you can directly get the record doing this:
`config_obj = self.env.ref('sbc_translation.nas_share_name')`

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +34,114 @@ class Correspondence(models.Model):
...
+            nas_share_name = (config_obj.search(
+                [('key', '=', 'sbc_translation.nas_share_name')])[0]).value
+
+            child = self.sponsorship_id.child_id
+            sponsor = self.sponsorship_id.partner_id
+            letter_name = "_".join(
+                (child.code, sponsor.ref, str(self.id)))
+            letter_name = ".".join((letter_name, 'pdf'))
+
+            nas_letters_store_path = (config_obj.search(
```
ecino: Here again you can use `self.env.ref`

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +34,114 @@ class Correspondence(models.Model):
...
+    # CRON Methods
+    ##############
+    @api.model
+    def check_local_translation_done(self):
+        reload(sys)
+        sys.setdefaultencoding('UTF8')
+        tc = translate_connector.TranslateConnect()
+        letters_to_update = tc.get_translated_letters()
+
+        if not letters_to_update == -1:
```
ecino: `if letters_to_update != -1` or simply `if letters_to_update`

sbc_translation/models/translate_connector.py
```diff
@@ -0,0 +1,159 @@
...
+            INNER JOIN translation tr\
+            ON trs.translation_id = tr.id\
+            INNER JOIN text txt\
+            ON tr.text_id = txt.id\
+            INNER JOIN language l\
+            ON txt.aim_lang_id = l.id\
+            WHERE tr.letter_odoo_id IS NOT NULL\
+            AND trs.status_id = 3"
+        )
+        return res if res else -1
```
ecino: selectAll returns an empty list if nothing is found. That would be more helpful than -1, as you can use a loop outside the function without testing the list contains something.

sbc_translation/models/correspondence.py
ecino: I think you can even remove the test, if `get_translated_letters` returns an empty list if nothing is returned -> the for loop will do nothing.

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +34,114 @@ class Correspondence(models.Model):
...
+                target_lang_id = self.env['res.lang.compassion'].search(
+                    [('code_iso', '=', tg_lang)]).id
+                # find the good text to writte
+                if tg_lang == 'eng':
+                    target_text = 'english_text'
+                elif tg_lang in correspondence.child_id.project_id\
+                        .country_id.spoken_lang_ids.mapped('code_iso'):
+                    target_text = 'translated_text'
+                else:
+                    raise AssertionError('The letter does not exist in odoo')
```
ecino: A better error message would be `letter correspondence.kit_identifier was translated in a wrong language: tg_lang`

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +34,114 @@ class Correspondence(models.Model):
...
+
+                # UPDATE Odoo Database
+                correspondence.write(
+                    {target_text: letter["text"],
+                     'state': 'Received in the system',
+                     'destination_language_id': target_lang_id})
+
+                # Send to GMC
+                if correspondence.direction == 'Supporter To Beneficiary':
+                    pdb.set_trace()
```
ecino: pdb!

sbc_translation/models/translate_connector.py
```diff
@@ -0,0 +1,159 @@
...
+    def __init__(self):
+        super(TranslateConnect, self).__init__(
+            'mysql_translate_host',
+            'mysql_translate_user',
+            'mysql_translate_pw',
+            'mysql_translate_db')
+
+        self.current_time = datetime.datetime.now()
+
+    def upsert_text(self, sponsorship, letter):
```
ecino: You could only pass `letter` as argument, and sponsosrhip would be `letter.sponsorship_id`

sbc_translation/models/correspondence.py
eicher31: How?

sbc_translation/models/translate_connector.py
```diff
@@ -0,0 +1,159 @@
...
+        self.letter_name = "_".join((child.code, sponsor.ref, str(letter.id)))
+        child_age = datetime.date.today().year - int(child.birthdate[:4])
+
+        src_lang_id = self.get_language_id(
+            letter.original_language_id.code_iso)
+
+        # Define the target language
+        child_lang = child.project_id.country_id.spoken_lang_ids
+        # eng, fra, deu, spa, por, ita
+        translate_language_ids = [1, 2, 3, 4, 5, 14]
```
ecino: You should never hardcode ids ! I would do :
`translate_language_ids = letter.env['res.lang.compassion'].search([('translatable', '=', True)]).ids`

sbc_translation/models/translate_connector.py
```diff
@@ -0,0 +1,159 @@
...
+            'text_type_id': text_type_id,
+        }
+        return self.upsert("text", vals)
+
+    def upsert_translation(self, text_id, letter):
+        """Push or update translation (db table) on local translate platform
+        """
+
+        vals = {
+            'file': self.letter_name + '.tif',
```
ecino: Should be `.pdf` ?

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +34,114 @@ class Correspondence(models.Model):
...
+                self.letter_image.with_context(
+                    bin_size=False).datas))
+            config_obj = self.env['ir.config_parameter']
+            nas_share_name = (config_obj.search(
+                [('key', '=', 'sbc_translation.nas_share_name')])[0]).value
+
+            child = self.sponsorship_id.child_id
+            sponsor = self.sponsorship_id.partner_id
+            letter_name = "_".join(
+                (child.code, sponsor.ref, str(self.id)))
```
ecino: I would setup the name at the beginning of the method and pass the name as a parameter to the function `upsert_text` to ensure the name is the same in the database! (if we change the code we don't have to change in the two methods).

sbc_translation/models/correspondence.py
```diff
@@ -22,22 +34,114 @@ class Correspondence(models.Model):
...
 
-    def alternative(self):
-        pass
+    def send_local_translate(self):
+        # Insert the letter in the mysql data base
+        tc = translate_connector.TranslateConnect()
+        # Send letter to local translate platform
+        text_id = tc.upsert_text(self.sponsorship_id, self)
+        translation_id = tc.upsert_translation(text_id, self)
+        tc.upsert_translation_status(translation_id)
```
ecino: I would write a `upsert_letter` method inside TranslateConnect that call those 3 methods, so that you don't have to pass the parameters each time, like you did for the `remove_letter` method.

sbc_translation/models/correspondence.py
```diff
@@ -112,36 +109,36 @@ def check_local_translation_done(self):
...
+                [('code_iso', '=', tg_lang)]).id
+            # find the good text to writte
+            if tg_lang == 'eng':
+                target_text = 'english_text'
+            elif tg_lang in correspondence.child_id.project_id\
+                    .country_id.spoken_lang_ids.mapped('code_iso'):
+                target_text = 'translated_text'
+            else:
+                raise AssertionError(
+                    'letter correspondence.kit_identifier was translated in\
```
ecino: `correspondence.kit_identifier` is a field, not a static text

ecino: But after thinking, this field is not set before we send it to gmc so you can replace by `correspondence.id`

sbc_translation/models/translate_connector.py
```diff
@@ -29,13 +29,13 @@ def __init__(self):
 
         self.current_time = datetime.datetime.now()
 
-    def upsert_text(self, sponsorship, letter):
+    def upsert_text(self, letter, letter_name):
         """Push or update text (db table) on local translate platform
         """
-        child = sponsorship.child_id
-        sponsor = sponsorship.partner_id
+        child = letter.sponsorship_id.child_id
```
ecino: I think child_id and partner_id are directly present in letter object. So you can do `letter.child_id` :wink:

ecino:
:+1:

PR #39, opened 2016-04-15
Refactor sbc_translation

ecino:
- Don't send e-mail to sponsor before letter is translated
- Allow setting letter translation from outside of odoo by adding method `update_translation`


MickSandoz:
Ok for me

PR #48, opened 2016-04-26
CO-15 When import failed, all files imported stay in their import location

MickSandoz:
When import success, files are stored on 'Done' + template_name folder


ecino:
Thanks :+1:

PR #54, opened 2016-05-03
CO-124 CO-125 Fix src_lang et dst_lang determination when translating…

ecino:
… letters

### DIFF
--- a/sbc_translation/models/correspondence.py
+++ b/sbc_translation/models/correspondence.py
@@ -76,28 +76,14 @@ class Correspondence(models.Model):
 
     @api.one
     def send_local_translate(self):
+        """
+        Sends the letter to the local translation platform.
+        :return: None
+        """
         child = self.sponsorship_id.child_id
 
         # Specify the src and dst language
-        src_lang_id = False
-        dst_lang_id = False
-        if self.direction == 'Supporter To Beneficiary':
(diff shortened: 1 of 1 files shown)


sbc_translation/models/correspondence.py
```diff
@@ -194,6 +152,88 @@ def update_translation(self, translate_lang, translate_text, translator):
...
+                raise Warning(_("Letter already sent to GMC cannot be "
+                                "translated! [%s]") % self.kit_identifier)
+
+            src_lang_id = self.original_language_id
+            child_langs = self.beneficiary_language_ids.filtered(
+                'translatable')
+            if not child_langs:
+                raise Warning(_("Child %s has no translatable language")
+                              % self.child_id.code)
+            dst_lang_id = child_langs[-1]
```
MickSandoz: if child speak english, it would not be preferable to set English here instead of the last language find int child_langs?
Or English is in all case the last element of child_langs?

ecino: Normally English is always in the child languages, however you are right, it is preferable to set it so that in case it is not set it won't fail. I will change this.

ecino: Corrected.

MickSandoz:
OK for me

PR #56, opened 2016-05-03
Import web letters in Odoo

MickSandoz:


sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,109 @@ def _get_smb_connection(self):
...
         """
         Call when a letter is set on web site:
             - add web letter to an import set with import letter config
               'Web letter'
         """
-        import_web_letter = self.search([
-            ('config_id.name', '=', 'Web Letter'),
-            ('state', '!=', 'done')])
+        child_id = None
+        if child_code:
```
ecino: child_code will always be None based on previous line

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,109 @@ def _get_smb_connection(self):
...
+            # Retrieve child code and find corresponding id
+            model_child = self.env['compassion.child'].search(
+                [('code', '=', child_code)], limit=1)
```
ecino: Instead of limiting search result, I would add the condition `state = 'P'` in the search parameters. Only one child with same code should be sponsored at a time.

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,109 @@ def _get_smb_connection(self):
...
+
+            child_id = model_child.id
+            logger.info("child id found {}".format(child_id))
```
ecino: I guess printing the child_code would be more useful as child_id will be None in that situation

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,109 @@ def _get_smb_connection(self):
...
+        else:
+            return False
+
+        sponsor_id = None
+        if sponsor_ref:
+            # Retrieve sponsor reference and find corresponding id
+            model_sponsor = self.env['res.partner'].search(
+                [('ref', '=', sponsor_ref)], limit=1)
```
ecino: Add condition `is_company = False`, because sometimes company partners can have same reference as the private person.

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,109 @@ def _get_smb_connection(self):
...
+            # TODO TO IMPROVE
+            template_id = model_template.id
+            logger.info("template id found {}".format(template_id))
+
+            template = model_template.browse(template_id)
+
+            if template:
+                logger.info("template found: {}".format(template.name))
+            else:
+                return False
```
ecino: If the template is not found, instead of aborting the import you could use the default template `self.env.ref('sbc_compassion.default_template')` (if I remember correctly)

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,109 @@ def _get_smb_connection(self):
...
+            logger.info("language id found {}".format(lang_id))
+
+        if template_name:
+            # Retrieve template name and find corresponding id
+            model_template = self.env['correspondence.template'].search(
+                [('name', '=', template_name)], limit=1)
+
+            template = None
+
+            # TODO TO IMPROVE
```
ecino: What is to improve?

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,109 @@ def _get_smb_connection(self):
             return SMBConnection(
                 SmbConfig.smb_user, SmbConfig.smb_pass, 'openerp', 'nas')
 
-    @api.model
-    def import_web_letter(self):
+    @api.one
+    def import_web_letter(self, child_code, sponsor_ref, language_name,
+                          original_text, template_name, pdf_letter=None):
```
ecino: Why is pdf_letter an optional argument if the import is aborted when no pdf is given ?

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,109 @@ def _get_smb_connection(self):
...
+        if pdf_letter:
+            file_pdf = BytesIO(base64.b64decode(pdf_letter))
+
+            filename = 'WEB_' + sponsor_ref + '_' + \
+                child_code + '_' + str(time.time())[:10] + '.pdf'
+
+            logger.info("Try to copy file {} !".format(filename))
+
+            # Copy file in attachment in the done letter folder
+            share_nas = 'it'
```
ecino: Those paths should be retrieved from the configuration like for the standard import.

sbc_email/models/import_letters_history.py
MickSandoz: old comment, i will remove it

sbc_email/models/import_letters_history.py
MickSandoz: Il will remove that, it was used during test phase.

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,109 @@ def _get_smb_connection(self):
             return SMBConnection(
                 SmbConfig.smb_user, SmbConfig.smb_pass, 'openerp', 'nas')
 
-    @api.model
-    def import_web_letter(self):
+    @api.one
```
ecino: Is this method called from the website ? Do you provide an id of an import from the caller or not ? If not, it is better to replace with `api.model`

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,109 @@ def _get_smb_connection(self):
...
-        # TODO Create import letter
+            return False
+
+        # analyze attachment to check template and create image preview
+        line_vals, document_vals = func.analyze_attachment(
+            self.env, base64.b64decode(pdf_letter), filename,
+            template)
+
+        for i in xrange(0, len(line_vals)):
+            line_vals[i]['import_id'] = self.id
```
ecino: Is `self.id` defined at that point ?

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,109 @@ def _get_smb_connection(self):
...
+            line_vals[i]['partner_id'] = sponsor_id
+            line_vals[i]['child_id'] = child_id
+            line_vals[i]['letter_language_id'] = lang_id
+            line_vals[i]['original_text'] = original_text
+            line_vals[i]['status'] = 'ok'
```
ecino: You could use `line_vals[i].update({..})` for setting all these values

sbc_email/models/import_letters_history.py
MickSandoz: Yes, an id is provided, import_letter id is given from the php code (if there is not import_letter with web letter config, one would be created from php code)

sbc_email/models/import_letters_history.py
MickSandoz: Yes it's defined, see comment before

sbc_email/models/import_letters_history.py
ecino: Ok, good then.

sbc_email/models/import_letters_history.py
ecino: :+1:

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,114 @@ def _get_smb_connection(self):
...
-            ('state', '!=', 'done')])
+        child_id = None
+        if child_code:
+            # Retrieve child code and find corresponding id
+            model_child = self.env['compassion.child'].search(
+                [('code', '=', child_code), ('state', '=', 'P')])
+
+            child_id = model_child.id
+            logger.info("child code found {}".format(model_child.code))
+        else:
```
ecino: Instead of `else`, I would put `if not child_id` to capture as well if child is not found given a wrong child_code.

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,114 @@ def _get_smb_connection(self):
...
-            ('config_id.name', '=', 'Web Letter'),
-            ('state', '!=', 'done')])
+        child_id = None
+        if child_code:
+            # Retrieve child code and find corresponding id
+            model_child = self.env['compassion.child'].search(
+                [('code', '=', child_code), ('state', '=', 'P')])
+
+            child_id = model_child.id
+            logger.info("child code found {}".format(model_child.code))
```
ecino: Remove logs (not needed for production)

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,114 @@ def _get_smb_connection(self):
...
+        sponsor_id = None
+        if sponsor_ref:
+            # Retrieve sponsor reference and find corresponding id
+            model_sponsor = self.env['res.partner'].search(
+                [('ref', '=', sponsor_ref),
+                 ('is_company', '=', False)])
+
+            sponsor_id = model_sponsor.id
+            logger.info("sponsor ref found {}".format(model_sponsor.ref))
+        else:
```
ecino: `if not sponsor_id`

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,114 @@ def _get_smb_connection(self):
...
+
+        sponsor_id = None
+        if sponsor_ref:
+            # Retrieve sponsor reference and find corresponding id
+            model_sponsor = self.env['res.partner'].search(
+                [('ref', '=', sponsor_ref),
+                 ('is_company', '=', False)])
+
+            sponsor_id = model_sponsor.id
+            logger.info("sponsor ref found {}".format(model_sponsor.ref))
```
ecino: remove log

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,114 @@ def _get_smb_connection(self):
...
+            return False
+
+        lang_id = None
+        if language_name:
+            # Retrieve language name and find corresponding id
+            model_lang = self.env['res.lang.compassion'].search(
+                [('name', '=', language_name)], limit=1)
+
+            lang_id = model_lang.id
+            logger.info(
```
ecino: remove log

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,114 @@ def _get_smb_connection(self):
...
+                "language name found {}".format(
+                    model_lang.display_name))
+
+        if template_name:
+            # Retrieve template name and find corresponding id
+            model_template = self.env['correspondence.template'].search(
+                [('name', '=', template_name)], limit=1)
+
+            template = None
```
ecino: I would put this before `if template_name`

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,114 @@ def _get_smb_connection(self):
...
+                [('name', '=', language_name)], limit=1)
+
+            lang_id = model_lang.id
+            logger.info(
+                "language name found {}".format(
+                    model_lang.display_name))
+
+        if template_name:
+            # Retrieve template name and find corresponding id
+            model_template = self.env['correspondence.template'].search(
```
ecino: Why don't you directly put `template = self.env[...].search(...)` ?

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,114 @@ def _get_smb_connection(self):
...
+
+            template = None
+
+            template_id = model_template.id
+            logger.info("template id found {}".format(template_id))
+
+            template = model_template.browse(template_id)
+
+            if template:
+                logger.info("template found: {}".format(template.name))
```
ecino: Remove log

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,114 @@ def _get_smb_connection(self):
...
+            template)
+
+        for i in xrange(0, len(line_vals)):
+            line_vals[i].update({
+                'import_id': self.id,
+                'partner_id': sponsor_id,
+                'child_id': child_id,
+                'letter_language_id': lang_id,
+                'original_text': original_text,
+                'status': 'ok'
```
ecino: Why do you force the status of the line ? If lang is not correctly given, or template is default, the status shouldn't be ok. I think it is better to remove this.

sbc_email/models/import_letters_history.py
MickSandoz: old code when I used cr, uuid,..  will correct that

sbc_email/models/import_letters_history.py
MickSandoz: I'm agree,.. I used it before in developmnent test phase...

sbc_email/models/import_letters_history.py
```diff
@@ -379,28 +380,101 @@ def _get_smb_connection(self):
...
+            file_pdf = BytesIO(base64.b64decode(pdf_letter))
+
+            filename = 'WEB_' + sponsor_ref + '_' + \
+                child_code + '_' + str(time.time())[:10] + '.pdf'
+
+            logger.info("Try to copy file {} !".format(filename))
 
-        if import_web_letter:
-            logger.info('Import letter with config Web Letter existence')
+            # Copy file in attachment in the done letter folder
```
ecino: I would move the samba code at the end of the function such that we only copy the file at the end after everything went well. In case of error, it doesn't lead to a file stored in done but not imported.

ecino:
:+1:

PR #57, opened 2016-05-10
Field "source" filled in import's letters

MickSandoz:
- add web config in data file import_config_templates.xml
-  set "source" field to "website" in import's letters


ecino:
:+1:

PR #60, opened 2016-05-17
Add deleting cascade in Mysql DB translate_test

eicher31:
Reformat some linked code
Before apply change in Odoo prod, update the DB translate_prod


ecino:
Can you explain what does this modification ?

eicher31:
Increase the reliability of deletion of a letter on the translation platform while simplifying access to the DB

PR #66, opened 2016-05-19
Save pictures attached to web letters

MickSandoz:


sbc_email/models/import_letters_history.py
```diff
@@ -151,6 +151,81 @@ def button_import(self):
...
             return super(ImportLettersHistory, self).button_import()
 
+    @api.multi
+    def button_save(self):
+        """
+        When saving the import_line as correspondences, move pdf file in the
+        done folder on the NAS and remove image attachment (Web letter case)
+        """
+        for letter in self:
+            if letter.config_id.name == "Web Letter":
```
ecino: Would be better to check `if letter.config_id == self.env.ref('sbc_email.web_letter')`

sbc_email/models/import_letters_history.py
```diff
@@ -151,6 +151,81 @@ def button_import(self):
 
             return super(ImportLettersHistory, self).button_import()
 
+    @api.multi
+    def button_save(self):
+        """
+        When saving the import_line as correspondences, move pdf file in the
+        done folder on the NAS and remove image attachment (Web letter case)
+        """
+        for letter in self:
```
ecino: Technically, would be better to call this `import` instead of `letter`.

sbc_email/models/import_letters_history.py
```diff
@@ -151,6 +151,81 @@ def button_import(self):
...
+                                    smb_conn.deleteFiles(share_nas,
+                                                         imported_letter_path +
+                                                         part_filename +
+                                                         image_ext)
+                                except Exception as inst:
+                                    logger.info('Failed to delete attached\
+                                                 image {}'.format(inst))
+
+            if letter.manual_import:
+                self._manage_all_imported_files()
```
ecino: Won't the code before already move the files of the manual import ?

sbc_email/models/import_letters_history.py
```diff
@@ -151,6 +151,81 @@ def button_import(self):
...
+
+                        # loop in import folder and find pdf corresponding to
+                        # this web letter and eventual attached image
+                        image_ext = None
+                        file_to_save = False
+                        for sharedFile in listPaths:
+                            ext = sharedFile.filename[-4:]
+                            if part_filename == sharedFile.filename[:-4]:
+                                if ext == '.pdf':
+                                    file_to_save = True
```
ecino: Should you add the check `self.is_in_list_letter` before putting `file_to_save` to True ?

sbc_email/models/import_letters_history.py
```diff
@@ -281,17 +357,20 @@ def _manage_all_imported_files(self):
...
-                        share_nas,
-                        imported_letter_path +
-                        sharedFile.filename,
-                        file_obj)
-                    file_obj.seek(0)
-                    self._copy_imported_to_done_letter(
-                        sharedFile.filename, file_obj, True)
+                    # when this is manual import we don't have to copy all
+                    # files, web letters are stock in the same folder...
+                    if not self.manual_import or self.is_in_list_letter(
```
ecino: Since you already did the check `not self.manual_import` at line 302 (before calling the function), I guess it becomes useless here or not ?

sbc_email/models/import_letters_history.py
```diff
@@ -303,10 +382,17 @@ def _manage_all_imported_files(self):
...
                         zip_file, 'r')
 
-                    # loop over files inside zip
+                    zip_to_remove = False
                     for f in zip_.namelist():
-                        self._copy_imported_to_done_letter(
-                            f, BytesIO(zip_.read(f)), False)
+                        # when this is manual import we are not sure that this
+                        # zip contains current letters treated
+                        if not self.manual_import or self.is_in_list_letter(f):
```
ecino: Is the manual_import check useless ?

sbc_email/models/import_letters_history.py
MickSandoz: No because code before is for import config web letter (test in lines 161, 162)
When it's a manual import, there is no import_config set, so code before is not execute

sbc_email/models/import_letters_history.py
MickSandoz: Not necessary in this case, the test made before is similary to test made in is_in_list_letter function

sbc_email/models/import_letters_history.py
MickSandoz: Not because this function is too called in line 304 with inverse case

sbc_email/models/import_letters_history.py
MickSandoz: No but I am not sure how python treats condition in "if" with  "or". In Java, if first condition is true the second is not checked. 
In this case it will avoid to execute is_in_list_letter function (not useless when whre are not in manual import)

ecino: It's the same in python, so that's good.

ecino:
:+1:

PR #73, opened 2016-06-01
Letters: check language of translated_text received

MickSandoz:
!!! WARNING !!!
install detectlanguage on server https://github.com/detectlanguage/detectlanguage-python

pip install detectlanguage

If the text is empty or not in the language received, send to translate to local platform

### DIFF
--- a/sbc_translation/models/correspondence.py
+++ b/sbc_translation/models/correspondence.py
@@ -10,13 +10,14 @@
 ##############################################################################
 import sys
 import base64
+import detectlanguage
 
 from io import BytesIO
 from openerp.exceptions import Warning
 
 from . import translate_connector
 
-from openerp import models, api, _
+from openerp import models, api, fields, _
 from openerp.tools.config import config
 
 from smb.SMBConnection import SMBConnection
@@ -31,6 +32,8 @@ class Correspondence(models.Model):
         """
 
     _inherit = 'correspondence'
+
+    has_valid_language = fields.Boolean(compute='_compute_has_valid_language')
 
     ##########################################################################
     #                              ORM METHODS                               #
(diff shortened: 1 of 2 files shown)


ecino:
Can you add a README file for the module (copy one of other module for reference)? You can then list the dependency in the README of the module, inside the `Installation` section. You can also list the dependency in the openerp file (see how it is done in other modules).

ecino:
I think an elegant way of doing this would be to add a computed field :

```
has_valid_language = fields.Boolean(compute='_compute_has_valid_language')
```
`_has_valid_language` could be like your method `is_writed_in_language`. This way the test could simply changed to : `or not letter.has_valid_language`.

sbc_translation/models/correspondence.py
```diff
@@ -68,12 +69,33 @@ def process_letter(self):
...
 
+    def is_writed_in_language(self, text, language_id):
+        """ Detect if text is writed in the language corresponding to the
+        language_id """
+        if text is not None and language_id is not None:
+            s = text.strip(' \t\n\r')
+            if s != "":
+                # find the language name of text argument
+                detectlanguage.configuration.api_key = \
+                    "ec1377a3d666c2ff8c60a252e2b9ec2f"
```
ecino: Not a good practice to put api keys in open source code :wink: 
Better read it from the configuration file (don't forget to specify in the README to add the key in the configuration file).

MickSandoz: Just I seen then when I pushed the code and I thought too that it would be good to change...

sbc_translation/models/correspondence.py
```diff
@@ -67,14 +70,38 @@ def process_letter(self):
         """ Called when B2S letter is Published. Check if translation is
          needed and upload to translation platform. """
         for letter in self:
+            self._compute_has_valid_language
```
ecino: No need to call the method. Accessing the field automatically calls the method.

MickSandoz: OK, I would like to ask you what means "compute"?
I have my answer ; -)

ecino:
Thanks :+1:

PR #80, opened 2016-06-13
Fix switzerland migrations after R4

ecino:
- sbc_email and sbc_translation adapations
- fix migration of child codes in GP
- fix child sync with gp


MickSandoz:
OK

PR #110, opened 2016-08-23
Add route for posting Christmas LSV/DD invoices

ecino:


mailing_christmas/controllers/generate_christmas_invoice.py
```diff
@@ -0,0 +1,134 @@
...
+from openerp import http, fields
+from openerp.tools import mod10r
+from openerp.http import request
+
+
+_logger = logging.getLogger(__name__)
+
+
+OK_RESULTS = {
+    'en_US': 'We successfully received your authorization. The amount will '
```
ecino: @davidwul We should validate the result messages. Are you ok with English and French? Then we can ask for German and Italian translation. Is it ok to say the debit will be done in the next 3 days?

philippe89: maybe realized

philippe89:
:+1:

PR #134, opened 2016-09-07
Christmas mass mailing fixes

ecino:
- Send e-mails in jobs
- Avoid unnecessary GP sync
- FIX birthdates of partners -> use real date fields with module partner_contact_birthdate

### DIFF
--- a/partner_sync_gp/model/partner_compassion.py
+++ b/partner_sync_gp/model/partner_compassion.py
@@ -11,6 +11,16 @@
 from openerp import api, fields, models
 from . import gp_connector
 
+SYNC_FIELDS = [
+    'abroad', 'birthdate', 'birthday_reminder', 'calendar', 'category_id',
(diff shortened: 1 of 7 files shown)


partner_sync_gp/model/partner_compassion.py
```diff
@@ -78,12 +88,18 @@ def write(self, vals):
...
-
         # We first call the original write method and update the records.
         result = super(ResPartner, self).write(vals)
 
+        # Don't synchronize if nothing relevant is changed.
+        for key in vals.iterkeys():
+            if key in SYNC_FIELDS:
+                break
+        else:
+            return result
```
philippe89: think line 98 and 99 should be one tab to the right

ecino: No it's a `for ... else`
http://psung.blogspot.ch/2007/12/for-else-in-python.html

philippe89: okay :+1:

PR #240, opened 2017-07-12
CO-1003 Update method to publish children on Wordpress website. 

jklay:
We now use direct XML-RPC calls to create children and don't upload pictures anymore, we use Cloudinary url's instead.

### DIFF
--- a/child_sync_wp/tools/wp_sync.py
+++ b/child_sync_wp/tools/wp_sync.py
@@ -8,12 +8,8 @@
 #    The licence is in the file __openerp__.py
 #
 ##############################################################################
-import base64
-import csv
 import logging
-import shutil
 
-from os import listdir, path, makedirs, remove
 from xmlrpclib import ServerProxy, SafeTransport, GzipDecodedResponse
 
 from odoo import _
@@ -21,12 +17,6 @@ from odoo.exceptions import UserError
 from odoo.tools import config
 
 _logger = logging.getLogger(__name__)
-
-try:
-    import pysftp
-    from wand.image import Image
-except ImportError:
-    _logger.warning("Please install wand and pysftp")
 
 
 # Solves XMLRPC Parse response problems by stripping response
@@ -67,13 +57,7 @@ class WPSync(object):
         host = config.get('wordpress_host')
         user = config.get('wordpress_user')
         password = config.get('wordpress_pwd')
-        sftp_host = config.get('wp_sftp_host')
-        sftp_user = config.get('wp_sftp_user')
-        sftp_pw = config.get('wp_sftp_pwd')
-        csv_path = config.get('wp_csv_path')
-        pic_path = config.get('wp_pictures_path')
-        if not (host and user and password and sftp_host and sftp_user and
-                sftp_pw and pic_path):
+        if not (host and user and password):
             raise UserError(
                 _("Please add configuration for Wordpress uploads")
             )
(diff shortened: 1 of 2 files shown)


child_sync_wp/tools/wp_sync.py
```diff
@@ -94,36 +94,51 @@ def test_xmlrpc(self):
...
-                self.sftp.put(pictures_folder + '/' + picture_file)
-        _logger.info(".... Pictures uploaded.")
-        shutil.rmtree(pictures_folder, ignore_errors=True)
         result = True
-
         try:
-            result = self.xmlrpc_server.child_import.addChildren(
-                self.user, self.pwd)
+            for child in children:
+                childarray = {}
```
ecino: You can directly put all values in the dictionary by doing
```
childarray = {
    'local_id': child.local_id,
    'number': child.local_id,
    ...
}
```

child_sync_wp/tools/wp_sync.py
```diff
@@ -94,36 +94,51 @@ def test_xmlrpc(self):
...
+                childarray['desc_it'] = child.desc_it
+                childarray['country'] = child.project_id.country_id.name
+                childarray['project'] = \
+                    child.project_id.description_fr
+                childarray['project_de'] = \
+                    child.project_id.description_de
+                childarray['project_it'] = \
+                    child.project_id.description_it
+                childarray['cloudinary_url'] = child.image_url
+                result = self.xmlrpc_server.child_import.addChild(
```
ecino: Is result still returning the number of children imported, or only True ? At the end, result should hold the number of successful imports to print it in the console. Like that, it will only contain the last result when you print it. So either you have to change to `result += self.xmlrpc_server.child_import.addChild(..)` or
```
if self.xmlrpc_server.child_import.addChild(...):
   result += 1
```

child_sync_wp/tools/wp_sync.py
```diff
@@ -94,36 +94,51 @@ def test_xmlrpc(self):
...
+                    self.user, self.pwd, childarray)
             if result:
                 _logger.info(
-                    "Child Upload on Wordpress finished: %s children imported "
+                    "Child Upload on Wordpress finished: %s child imported "
```
ecino: better to keep children and print the total imports

child_sync_wp/tools/wp_sync.py
ecino: Convention for naming variables is to put `_` where spaces would be required. So this would be `child_array` or better `child_values`

ecino (changes requested):
Great simplification! Small code improvements can be done. (you can use git amend commit and git force push to correct your pull request)

ecino:
Travis build failed : 
```
child_sync_wp/tools/wp_sync.py:11:1: F401 'base64' imported but unused
child_sync_wp/tools/wp_sync.py:12:1: F401 'csv' imported but unused
child_sync_wp/tools/wp_sync.py:14:1: F401 'shutil' imported but unused
child_sync_wp/tools/wp_sync.py:16:1: F401 'os.listdir' imported but unused
child_sync_wp/tools/wp_sync.py:16:1: F401 'os.makedirs' imported but unused
child_sync_wp/tools/wp_sync.py:16:1: F401 'os.remove' imported but unused
child_sync_wp/tools/wp_sync.py:16:1: F401 'os.path' imported but unused
child_sync_wp/tools/wp_sync.py:27:5: F401 'wand.image.Image' imported but unused
```
You can remove all useseless imports in wp_sync.py now that a lot of the code has been removed. Once this is done, we can merge.

ecino (approved):

PR #241, opened 2017-07-12
CO-678 Alert when we try to add a partner who already exist

nfluckiger:


partner_compassion/models/partner_compassion.py
```diff
@@ -76,6 +76,10 @@ def _get_receipt_types(self):
         required=True,
         help='Delivery preference for Child photo')
 
+    listehomonyme = fields.Many2many(
```
ecino: Odoo has convention for naming fields. They should be with `_` where spaces would be present, they should be in English and there as well conventions for Many2one fields should en by `_id` and One2many and Many2many should end with `_ids`. I propose that you rename this field to `partner_duplicate_ids`

partner_compassion/models/partner_compassion.py
```diff
@@ -76,6 +76,10 @@ def _get_receipt_types(self):
...
+        'res.partner', 'homonyme', 'partner_id',
```
ecino: Change `homonyme` to `res_partner_duplicates`

partner_compassion/models/partner_compassion.py
```diff
@@ -76,6 +76,10 @@ def _get_receipt_types(self):
...
+        'partner_id_homonyme', readonly=True)
```
ecino: Change `partner_id_homonyme` to `duplicate_id`

partner_compassion/models/partner_compassion.py
```diff
@@ -183,6 +187,29 @@ def name_search(self, name, args=None, operator='ilike', limit=80):
         return self.browse(to_ret_ids).name_get()
 
     ##########################################################################
+    #                             ONCHANGE METHODS                           #
+    ##########################################################################
+
+    @api.onchange('lastname', 'birthdate_date', 'firstname')
```
ecino: Your filter is too restrictive. As we don't always put the birthdate info, I would prefer that you remove the check of the birthdate but add a check for the `zip` field, to check if the partner lives in the same city. You can as well check for the e-mail address which is a good indicator.

partner_compassion/models/partner_compassion.py
```diff
@@ -183,6 +187,29 @@ def name_search(self, name, args=None, operator='ilike', limit=80):
...
+    def _onchange_partner(self):
+        db_info = self.search([('firstname', '=', self.firstname),
+                               ('lastname', '=', self.lastname),
+                               ('birthdate_date', '=', self.birthdate_date)])
+        list_hom = []
+        for inf in db_info:
+            list_hom.append((4, inf.id))
+        self.listehomonyme = list_hom
+        self.homonymeformat = list_hom
```
ecino: You can change all the code of the method up to here by :
```
self.partner_duplicate_ids = self.search([
    '|', ('email', '=', self.email),
    '&', '&', ('firstname', 'ilike', self.firstname),
    ('lastname', 'ilike', self.lastname),
    ('zip', '=', self.zip)
])
```
The search filters is equivalent to saying "All partner with same e-mail OR (firstname and lastname like AND zip equal).
You can directly assign the result to your Many2many field without having to create a list of (4, id) tuples. The (4, id) tuple is only useful when calling the method `write` on a recordset. When you have a single record (which is the case in api.onchange methods), you can use assignation like `self.name = 'Test name'` or put recordset in relational fields.

partner_compassion/models/partner_compassion.py
```diff
@@ -183,6 +187,29 @@ def name_search(self, name, args=None, operator='ilike', limit=80):
...
+        if db_info:
```
ecino: `if self.partner_duplicate_ids:`

partner_compassion/models/partner_compassion.py
```diff
@@ -183,6 +187,29 @@ def name_search(self, name, args=None, operator='ilike', limit=80):
...
+            return {
+                     'warning': {'title': "Check-Double", 'message': "The "
```
ecino: Change title to `_("Possible existing partners found")`. Using the `_()` function will translate the string to the user language.

partner_compassion/models/partner_compassion.py
```diff
@@ -183,6 +187,29 @@ def name_search(self, name, args=None, operator='ilike', limit=80):
...
+                                 "partner you want to dd may already be in "
```
ecino: Change message to : `_('The partner you want to add may already exist. Please use the "Check duplicates" button to review it.')`

partner_compassion/models/partner_compassion.py
```diff
@@ -250,3 +277,26 @@ def _can_geocode(self):
         if 'no_geocode' in self.env.context:
             return False
         return super(ResPartner, self)._can_geocode()
+
+    @api.multi
+    def openduplication(self):
```
ecino: rename to `open_duplicates`

partner_compassion/models/partner_compassion.py
```diff
@@ -250,3 +277,26 @@ def _can_geocode(self):
...
+        db_info = self.search([('firstname', '=', self.firstname),
+                               ('lastname', '=', self.lastname),
+                               ('birthdate_date', '=', self.birthdate_date),
+                               ('id', '!=', self.id)])
+        list_hom = []
+        for inf in db_info:
+            list_hom.append((4, inf.id))
```
ecino: You don't need all this code. Here you can just use `self.partner_duplicate_ids` which was set in the `onchange` method.
Here is a pythonic shorcut :
```
duplicate_list = [(4, id) for id in self.partner_duplicate_ids]
```
Here is useful documentation about [list comprehensions](http://treyhunner.com/2015/12/python-list-comprehensions-now-in-color/)

partner_compassion/views/partner_compassion_view.xml
```diff
@@ -55,6 +55,13 @@
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="sbc_compassion.view_partner_form_sbc"/>
         <field name="arch" type="xml">
+            <!-- Add a Check double button on the top of the view-->
+            <xpath expr="//form/*[1]" position="before">
+                <header>
+                     <field name="listehomonyme" attrs="{'invisible':True}"/>
+                    <button type='object' name='openduplication' string="Check Double" attrs="{'invisible': [('listehomonyme','=', [])]}"/>
```
ecino: `string="Check duplicates"`

partner_compassion/wizards/partner_check_double.py
```diff
@@ -0,0 +1,39 @@
...
+##############################################################################
+
+from odoo import api, fields, models, _
+
+
+class PartnerCheckDouble(models.TransientModel):
+    _name = "res.partner.check.double"
+
+    newpartner_id = fields.Many2one('res.partner')
+    mergable = fields.Many2many(
```
ecino: Change field name to `mergeable_partner_ids`

ecino (changes requested):

partner_compassion/models/partner_compassion.py
```diff
@@ -114,6 +118,14 @@ def get_unreconciled_amount(self):
...
     @api.model
     def create(self, vals):
+        duplicate = self.search([
+                '|', '&', ('email', '=', vals.get('email')),
+                          ('email', '!=', False),
+                '&', '&', ('firstname', 'ilike', vals.get('firstname')),
+                          ('lastname', 'ilike', vals.get('lastname')),
+                          ('zip', '=', vals.get('zip'))])
+        duplicate_ids = [(4, itm.id) for itm in duplicate]
+        vals.update({'partner_duplicate_ids': duplicate_ids})
```
ecino: Are you obliged to do this? Isn't it working already if field is not readonly?

partner_compassion/views/partner_check_double.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="partner_check_double_wizards" model="ir.ui.view">
+        <field name="name">res.partner.check.double.view</field>
+        <field name="model">res.partner.check.double</field>
+        <field name="arch" type="xml">
+            <form>
+                <field name="mergeable_partner_ids"/>
+                <field name="selected_merge_partner_id" domain="[('id', 'in', mergeable_partner_ids[0][2])]"/>
+                <button type='object' name='keep' string="Conserver" />
```
ecino: Woops, strings must always be in English! Change with `Keep`

partner_compassion/views/partner_check_double.xml
```diff
@@ -0,0 +1,15 @@
...
+                <button type='object' name='mergewith' string="Fusionner"/>
```
ecino: `Merge`

partner_compassion/wizards/partner_check_double.py
```diff
@@ -0,0 +1,39 @@
...
+from odoo import api, fields, models, _
+
+
+class PartnerCheckDouble(models.TransientModel):
+    _name = "res.partner.check.double"
+
+    newpartner_id = fields.Many2one('res.partner')
+    mergeable_partner_ids = fields.Many2many(
+        'res.partner', 'res_partner_check_double_relation', 'wizard_id',
+        'partner_id', readonly=True)
```
ecino: add attribute `string="Similar partners found"` (it will be displayed as field name in the view)

partner_compassion/wizards/partner_check_double.py
```diff
@@ -0,0 +1,39 @@
...
+    selected_merge_partner_id = fields.Many2one('res.partner')
```
ecino: Add attribute `fields.Many2one('res.partner', 'Merge with')` (it will be the field name in the view)

ecino (changes requested):

partner_compassion/wizards/partner_check_double.py
```diff
@@ -0,0 +1,39 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2016 Compassion CH (http://www.compassion.ch)
```
ecino: 2017 you mean 💃

partner_compassion/models/partner_compassion.py
nfluckiger: If i put the filed not readonly in the .py and readonly in the view it doesn't work, so that the best option

ecino (approved):

PR #244, opened 2017-07-27
10.0 Update church sponsorship

nfluckiger:
the commit CO-819 just remove a blank line

### DIFF
--- a/partner_compassion/models/partner_compassion.py
+++ b/partner_compassion/models/partner_compassion.py
@@ -79,12 +79,14 @@ class ResPartner(models.Model):
     partner_duplicate_ids = fields.Many2many(
         'res.partner', 'res_partner_duplicates', 'partner_id',
         'duplicate_id', readonly=True)
+    church_member_count = fields.Integer(compute='_is_church',
+                                         readonly=True)
 
     ##########################################################################
     #                             FIELDS METHODS                             #
     ##########################################################################
     @api.multi
-    @api.depends('category_id')
+    @api.depends('category_id', 'member_ids', 'church_member_count')
     def _is_church(self):
         """ Tell if the given Partners are Church Partners
             (by looking at their categories). """
@@ -96,7 +98,10 @@ class ResPartner(models.Model):
             is_church = False
             if church_category in record.category_id:
                 is_church = True
+
+            record.church_member_count = len(record.member_ids.ids)
             record.is_church = is_church
+
 
     @api.multi
     def get_unreconciled_amount(self):
--- a/partner_compassion/models/partner_compassion.py
+++ b/partner_compassion/models/partner_compassion.py
@@ -79,14 +79,13 @@ class ResPartner(models.Model):
     partner_duplicate_ids = fields.Many2many(
         'res.partner', 'res_partner_duplicates', 'partner_id',
         'duplicate_id', readonly=True)
-    church_member_count = fields.Integer(compute='_is_church',
-                                         readonly=True)
+    church_member_count = fields.Integer(compute='_is_church')
 
     ##########################################################################
     #                             FIELDS METHODS                             #
     ##########################################################################
     @api.multi
-    @api.depends('category_id', 'member_ids', 'church_member_count')
+    @api.depends('category_id', 'member_ids')
     def _is_church(self):
         """ Tell if the given Partners are Church Partners
             (by looking at their categories). """
@@ -99,9 +98,8 @@ class ResPartner(models.Model):
             if church_category in record.category_id:
                 is_church = True
 
-            record.church_member_count = len(record.member_ids.ids)
+            record.church_member_count = len(record.member_ids)
             record.is_church = is_church
-
 
     @api.multi
     def get_unreconciled_amount(self):
@@ -117,6 +115,36 @@ class ResPartner(models.Model):
         for move_line in move_line_ids:
             res += move_line.credit
         return res
+
+    @api.multi
+    def _compute_children(self):
+        for partner in self:
+            partner.number_children = len(partner.sponsored_child_ids)
+            if partner.is_church:
+                partner.number_children += len(partner.mapped(
+                    'member_ids.sponsored_child_ids'))
+
+    @api.multi
+    @api.depends('category_id')
+    def _compute_has_sponsorships(self):
+        """
+        A partner is sponsor if he is correspondent of at least one
+        sponsorship.
+        """
+        for partner in self:
+            if partner.is_church:
+                nb_sponsorships = self.env['recurring.contract'].search_count(
+                    ['|', '|',
+                        ('correspondant_id', 'in', partner.member_ids.ids),
+                        ('correspondant_id', '=', partner.id), '|',
+                        ('partner_id', '=', partner.id),
+                        ('partner_id', 'in', partner.member_ids.ids),
+                        ('type', 'like', 'S')
+                     ])
+                partner.has_sponsorships = nb_sponsorships
+                partner.number_sponsorships = nb_sponsorships
+            else:
+                super(ResPartner, self)._compute_has_sponsorships()
 
     ##########################################################################
     #                              ORM METHODS                               #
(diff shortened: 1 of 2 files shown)


partner_compassion/models/partner_compassion.py
```diff
@@ -79,12 +79,14 @@ def _get_receipt_types(self):
     partner_duplicate_ids = fields.Many2many(
         'res.partner', 'res_partner_duplicates', 'partner_id',
         'duplicate_id', readonly=True)
+    church_member_count = fields.Integer(compute='_is_church',
+                                         readonly=True)
```
ecino: compute fields are readonly by default so it's not necessary.

partner_compassion/models/partner_compassion.py
```diff
@@ -79,12 +79,14 @@ def _get_receipt_types(self):
...
 
     ##########################################################################
     #                             FIELDS METHODS                             #
     ##########################################################################
     @api.multi
-    @api.depends('category_id')
+    @api.depends('category_id', 'member_ids', 'church_member_count')
```
ecino: Remove `church_member_count`. You cannot make your computed field depend on itself, it will produce bad behaviour.

partner_compassion/models/partner_compassion.py
```diff
@@ -96,8 +98,11 @@ def _is_church(self):
             is_church = False
             if church_category in record.category_id:
                 is_church = True
+
+            record.church_member_count = len(record.member_ids.ids)
```
ecino: remove `.ids`

partner_compassion/models/partner_compassion.py
```diff
@@ -96,8 +98,11 @@ def _is_church(self):
...
             record.is_church = is_church
 
+
```
ecino: This line is not needed

partner_compassion/views/partner_compassion_view.xml
```diff
@@ -96,7 +96,8 @@
 
             <!-- Add the church members list in a church partner -->
             <xpath expr="//page[2]" position="after">
-                <page string="Members" attrs="{'invisible': [('is_church','=',False), ('member_ids', '=', [])]}" autofocus="autofocus">
+                <page string="Members" attrs="{'invisible': ['|',('is_church','=',False), ('member_ids', '=', [])]}" autofocus="autofocus">Member(
+                    <field name="church_member_count"/>)
```
ecino: This is not the way of displaying info. Instead of directly writing strings `Members()` outside of fields you should rather add
```
<label for="church_member_count"/> -- this will display the label for the field
<field name="church_member_count"/>
```

ecino (changes requested):

partner_compassion/models/partner_compassion.py
```diff
@@ -23,6 +23,32 @@ class ResPartner(models.Model):
...
+        sponsorship.
+        """
+        for partner in self:
+            if partner.is_church:
+                partner.has_sponsorships = self.env[
+                    'recurring.contract'].search_count([
+                        '|',
+                        ('partner_id', 'in', partner.member_ids.ids),
+                        ('correspondant_id', 'in', partner.member_ids.ids),
+                        ('type', 'like', 'S')
```
ecino: You must as well include the direct sponsorships of the church, by adding the condition (with OR operator) `('partner_id', '=', partner.id)` and same for correspondant_id

partner_compassion/models/partner_compassion.py
```diff
@@ -23,6 +23,32 @@ class ResPartner(models.Model):
...
+                    ])
+                partner.number_sponsorships = self.env[
```
ecino: I think you can use the same search for number_sponsorships and has_sponsorships. Just do
```
nb_sponsorships = search_count([...])
partner.nb_sponsorships = nb_sponsorships
partner.has_sponsorships = nb_sponsorships
```
You can as well correct it in compassion-modules where I believe it's as well duplicated for nothing

ecino (changes requested):
Does it work as well for the number of children count (and the sponsored children button) ?

partner_compassion/models/partner_compassion.py
```diff
@@ -23,6 +23,79 @@ class ResPartner(models.Model):
...
+                'domain': ['|', '|',
+                           ('correspondant_id', '=', self.member_ids.ids),
+                           ('correspondant_id', '=', self.id), '|',
+                           ('partner_id', '=', self.id),
+                           ('partner_id', 'in', self.member_ids.ids)],
+                'context': self.with_context({
+                    'default_type': 'S'}).env.context,
+            }
+        else:
+            super(ResPartner, self).open_contracts()
```
ecino: You should `return super(..)`

partner_compassion/models/partner_compassion.py
```diff
@@ -23,6 +23,79 @@ class ResPartner(models.Model):
...
+                'res_model': 'compassion.child',
+                'view_type': 'form',
+                'view_mode': 'tree,form',
+                'domain': ['|', ('sponsor_id', '=', self.id),
+                           ('sponsor_id', 'in', self.member_ids.ids)],
+                'context': self.with_context({
+                    'default_type': 'S'}).env.context,
+            }
+        else:
+            super(ResPartner, self).open_sponsored_children()
```
ecino: Return super

partner_compassion/models/partner_compassion.py
```diff
@@ -23,6 +23,79 @@ class ResPartner(models.Model):
     """
     _inherit = 'res.partner'
 
+    @api.multi
+    def open_sponsored_children(self):
```
ecino: Please move the two `open` methods in the `VIEW CALLBACKS` section

partner_compassion/models/partner_compassion.py
```diff
@@ -23,6 +23,79 @@ class ResPartner(models.Model):
...
+                           ('partner_id', '=', self.id),
+                           ('partner_id', 'in', self.member_ids.ids)],
+                'context': self.with_context({
+                    'default_type': 'S'}).env.context,
+            }
+        else:
+            super(ResPartner, self).open_contracts()
+
+    @api.multi
+    def _compute_children(self):
```
ecino: Please move the `_compute` methods inside the `FIELDS METHODS` section

ecino (changes requested):

ecino:
Also watch for flake8 errors (you can view them in the PYLINT results of the travis build :
```
partner_compassion/models/partner_compassion.py:76:18: E111 indentation is not a multiple of four
partner_compassion/models/partner_compassion.py:77:18: E111 indentation is not a multiple of four
```

partner_compassion/models/partner_compassion.py
```diff
@@ -113,6 +116,39 @@ def get_unreconciled_amount(self):
...
+        church = None
+        total = 0
+        for partner in self:
+            if partner.is_church:
+                church = partner
+            else:
+                partner.number_children = len(partner.sponsored_child_ids)
+                total += len(partner.sponsored_child_ids)
+        if church:
```
ecino: Your way of doing it is wrong: If many churches are in `self`, you will only update the last one. Why not doing this directly in the for loop? Here you are adding all the the members of all churches and putting the result in the last church.

partner_compassion/models/partner_compassion.py
```diff
@@ -113,6 +116,39 @@ def get_unreconciled_amount(self):
...
+        sponsorship.
+        """
+        for partner in self:
+            if partner.is_church:
+                nb_sponsorships = self.env['recurring.contract'].search_count(
+                    ['|', '|',
+                        ('correspondant_id', 'in', partner.member_ids.ids),
+                        ('correspondant_id', '=', partner.id), '|',
+                        ('partner_id', '=', partner.id),
+                        ('partner_id', 'in', partner.member_ids.ids)])
```
ecino: You missed the last ('type', 'like', 'S') condition

partner_compassion/models/partner_compassion.py
```diff
@@ -269,6 +305,46 @@ def get_lang_from_phone_number(self, phone):
...
+        if self.is_church:
+            return {
+                'type': 'ir.actions.act_window',
+                'name': 'Children',
+                'res_model': 'compassion.child',
+                'view_type': 'form',
+                'view_mode': 'tree,form',
+                'domain': ['|', ('sponsor_id', '=', self.id),
+                           ('sponsor_id', 'in', self.member_ids.ids)],
+                'context': self.with_context({
```
ecino: self.env.context is enough here (no need to add `default_type` in context.

ecino (changes requested):

ecino (approved):

PR #248, opened 2017-08-07
Odoo 8 to Odoo 10 change

nfluckiger:
Modification for the migration of sbc_switzerland


sbc_switzerland/models/import_letters_history.py
```diff
@@ -31,6 +31,7 @@
     from smb.smb_structs import OperationFailure
     from pyPdf import PdfFileWriter, PdfFileReader
     from pyPdf.pdf import PageObject
+    from openerp.addons.connector.queue.job import job, related_action
```
ecino: `from odoo.addons.queue_job.job import job, related_action`

PR #249, opened 2017-08-08
CO-852 - Direct Debit not in 1098

maxime-beck:


ecino:
I will check with Rachel if it's still ok. I'm in doubt now 🤔

ecino:
So in fact this change is no longer needed. You can replace it by adding another line `u'CRÉDIT GROUPÉ BVR'` in the list of lsv_dd_strings. Sorry for the confusion.

PR #256, opened 2017-08-15
Fix Number of Sponsorships

maxime-beck:


davidwul (changes requested):
Actually, it's the opposite. sponsorship active = state active and sponsorship_disbursed is state=active and hold_cdsp_funds = false (hold means don't disbursed)

ecino:
As it is now, I don't see any difference in the result, just another way of having the same list.

maxime-beck:
@ecino The difference resides in the fact that apparently, the hold_cdsp_funds of ICPs may be null when it does not contains life cycles[1]. Because of this, the first solution's results were wrong.

To prevent that, maybe we could set the default value of all the booleans life cycles relational fields to false?

---
[1] https://github.com/CompassionCH/compassion-modules/blob/10.0/child_compassion/models/project_compassion.py#L280

ecino (approved):
In that case I think your change is good, we don't need to set default values. @davidwul you can merge if ok for you.

davidwul (approved):

PR #262, opened 2017-09-29
10.0 bug fixes

ecino:


eicher31 (approved):
That's look good, at least on the python side! About the  Odoo's syntax I can't say more...

PR #271, opened 2017-10-16
CO-1074 / CO-1075

eicher31:


mass_mailing_switzerland/models/import_letters_history.py
```diff
@@ -78,10 +80,14 @@ def import_web_letter(self, child_code, sponsor_ref,
                 sponsor_id = model_sponsor.id
 
             # Check if a sponsorship exists
+            current_date = datetime.now()
+            limit_date = current_date + timedelta(weeks=12)
             self.env['recurring.contract'].search([
                 ('child_id', '=', child_id),
                 ('correspondant_id', '=', sponsor_id),
-                ('is_active', '=', True)], limit=1).ensure_one()
+                ('|', ('end_date', '=', False),
```
ecino: I don't know if it's the same or not but usually I don't put the "or" condition inside a tuple. It would be simply
`'|', ('end_date', '=', False), ('end_date', '<=', ...)]`.
The operator is automatically applied between the next two tuples.

ecino (commented):
You have as well some pep8 errors.

ecino (approved):

PR #273, opened 2017-10-18
CO-1175 The wrong module name was used for the differents languages

cmtCousteau:


ecino (changes requested):
Waiting for the code that has to be moved from compassion-modules to compassion-switzerland.

sbc_switzerland/models/correspondence_template.py
```diff
@@ -0,0 +1,26 @@
+from odoo import models, api, fields
```
ecino, line 12: Please add a header (like in other files, with your name as author)

ecino (changes requested):

PR #293, opened 2017-11-21
CO-1258 UPDATE

eicher31:
The methode create_from_wordpress update now the field reference with the postfinance transaction number


ecino (changes requested):
You have an indentation error (see flake8 error)

ecino (approved):

PR #294, opened 2017-11-23
CO-869 Add CRON to cleanup mail_message in database

ecino:


cleanup_switzerland/models/database_cleanup.py
```diff
@@ -0,0 +1,132 @@
...
+                _logger.info(
+                    '{} mail_message for invalid {} objects removed'.format(
+                        count, model)
+                )
+            except ProgrammingError:
+                self.env.cr.rollback()
+
+        # Delete old messages for old objects not useful
+        today = date.today()
+        limit = today - relativedelta(years=1)
```
eicher31, line 74: For donations certificate maybe the accounting team need to see more than just one year in the past?

eicher31 (approved):

cleanup_switzerland/models/database_cleanup.py
ecino: This is only cleaning the messages, meaning the history of the changes of status for instance. We will still be able to see the invoices, the accounting entries and everything needed for donations certificate.

PR #309, opened 2017-12-19
CO-1324 - Now check for parent mandate

cmtCousteau:


sponsorship_switzerland/models/contracts.py
```diff
@@ -65,6 +65,10 @@ def _compute_has_mandate(self):
             count = self.env['account.banking.mandate'].search_count([
                 ('partner_id', '=', contract.partner_id.id),
                 ('state', '=', 'valid')])
+            if contract.partner_id.parent_id.id:
```
ecino: if contract.partner_id.parent_id:

ecino (changes requested):
rebase

ecino (approved):

PR #318, opened 2018-01-18
New Tax Receipt + SUB proposition communications

ecino:


ecino:
I merge,  @eicher31 you can still review it in case you find something afterwards we will correct it.

PR #359, opened 2018-03-02
Add statistics charts for sponsorships

popod:


sponsorship_switzerland/__manifest__.py
```diff
@@ -42,6 +42,7 @@
         'account_statement_completion',
         'l10n_ch_lsv_dd',
         'gift_compassion',
+        'recurring_contract'
```
ecino: This is not needed. The other modules listed above already depends on module `recurring_contract`

sponsorship_switzerland/reports/sponsorships_evolution_months_report.py
```diff
@@ -0,0 +1,45 @@
...
+        self.env.cr.execute("""
+            CREATE OR REPLACE VIEW sponsorships_evolution_months_report AS
+            SELECT
+              sub.start_date AS start_date,
+              ROW_NUMBER() OVER (ORDER BY (SELECT 100)) AS id,
+              SUM(sub.total) over (ORDER BY sub.start_date)
+                AS cumulative_sponsorships
+            FROM (
+              SELECT
+                to_char(date_trunc('month', rc.start_date),'YYYY.MM')
```
ecino: Instead of the field start_date, we want to use the field `activation_date` which says when the sponsor paid the first time the sponsorship and it's the date where we consider the sponsorship is active

sponsorship_switzerland/reports/sponsorships_evolution_months_report.py
```diff
@@ -0,0 +1,45 @@
...
+                AS start_date, COUNT(rc.start_date) AS total
+              FROM recurring_contract AS rc
+              WHERE rc.state = 'active'
```
ecino: Replace this filter by `activation_date is not null` because we should include all sponsorship that are no longer active now but were active in the past.

ecino (changes requested):
Something missing that will probably be hard to implement: we must take into consideration when the sponsorship has ended. For instance we look at the month of January 2016 : the cumulative count here should be : `All sponsorships with activation_date < 01.2016 and end_date null or > 01.2016`

sponsorship_switzerland/__manifest__.py
popod: Okey, done.

sponsorship_switzerland/reports/sponsorships_evolution_months_report.py
popod: Yes, true. I've done this change.

sponsorship_switzerland/reports/sponsorships_evolution_months_report.py
popod: yes right.

PR #375, opened 2018-03-13
CO-1523 Add option to remove PP box on communications

popod:


report_compassion/report/partner_communication.xml
```diff
@@ -227,7 +229,7 @@
             </div>
             <div id="letter">
                 <div id="address" class="right">
-                    <div id="pp">
+                    <div id="pp" t-attf-style="visibility: {{ 'visible' if display_pp else 'hidden' }};">
```
ecino: Here instead of your t-attf-style attribute you can simply put `t-if="display_pp"`

ecino (changes requested):

PR #388, opened 2018-03-22
CO-1545 Lors du lettrage: Trier par partenaire-mchanging the display …

vwvw:
…order of bank account statements during the reconciliation process from partner_id to lastname, firstname

### DIFF
--- a/account_reconcile_compassion/models/bank_statement.py
+++ b/account_reconcile_compassion/models/bank_statement.py
@@ -42,85 +42,25 @@ class AccountStatement(models.Model):
 
     @api.multi
     def reconciliation_widget_preprocess(self):
-        """ Overwrite with slight changes
-            odoo/addons/account/models/accound_bank_statement.py
-
-            Get statement lines of the specified statements or all
-            unreconciled statement lines and try to automatically reconcile
-            them / find them a partner.
-            Return ids of statement lines left to reconcile and other data
-            for the reconciliation widget.
         """
-        statements = self
-        # NB : The field account_id can be used at the statement line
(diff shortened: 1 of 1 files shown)


account_reconcile_compassion/models/bank_statement.py
```diff
@@ -38,3 +39,88 @@ def _default_name(self):
...
+            sql_query += ' ORDER BY partner.lastname, partner.firstname, ' \
+                         'partner.id'
+            self.env.cr.execute(sql_query, params)
+            results = self.env.cr.dictfetchall()
+            st_line = self.env['account.bank.statement.line']
+            for line in results:
+                st_line.browse(line.get('id')).write(
+                    {'partner_id': line.get('partner_id')})
+        return {
+            'st_lines_ids': lines_left.ids,
```
ecino: Could you just instead use the super method in order to avoid copying all the code? This way if something changes, we don't override everything. I think of something like :
```
result = super(..., self).reconciliation_widget_preprocess()
line_ids = result['st_lines_ids']
... reorder process
result['st_lines_ids'] = new_order_ids
return result
```

ecino (changes requested):

account_reconcile_compassion/models/bank_statement.py
vwvw: Missed this opportunity. Should be much better now !

ecino (approved):

ecino:
account_reconcile_compassion/models/bank_statement.py:57:64: W291 trailing whitespace

PR #393, opened 2018-03-27
CO-1566 Muskathlon reports

popod:


muskathlon/README.rst
```diff
@@ -0,0 +1,30 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Muskathlon Switzerland
+====================
```
ecino: Line must be equal length of the title

muskathlon/__manifest__.py
```diff
@@ -0,0 +1,45 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+# pylint: disable=C8101
+{
+    'name': 'Muskathlon',
+    'version': '10.0.0.0.0',
```
ecino: 10.0.1.0.0

ecino (changes requested):
Please move all muskathlon fields that are declared in module partner_communication_switzerland in your module muskathlon.

muskathlon/reports/muskathlon_report.py
```diff
@@ -0,0 +1,119 @@
...
+              LEFT JOIN res_partner AS rp ON rp.id = ai.partner_id
+              LEFT JOIN res_currency AS rcu ON rcu.id = ai.currency_id
+              LEFT JOIN crm_event_compassion AS cec ON ail.event_id = cec.id
+              LEFT JOIN account_invoice_account_move_line_rel AS aiamlr
+                ON ai.id = aiamlr.account_invoice_id
+              LEFT JOIN account_move_line AS aml
+                ON aml.id = aiamlr.account_move_line_id
+              WHERE ail.state = 'paid' AND cec.muskathlon_event_id IS NOT NULL
+            )
+        """.format(table=self._table))
```
ecino: format should not be used with SQL queries because it can lead to SQL injections. You should use instead %s replacement character and use pass a list parameters to the execute function `self.env.cr.execute(query, [self._table])`

ecino (changes requested):
You should as well add an access rights file for the new model muskathlon.report (otherwise only the administrator will have access to the report)

muskathlon/reports/muskathlon_report.py
popod: Ok, done ;)

ecino (approved):

PR #419, opened 2018-04-18
Muskathlon website

popod:
Need to be done:
- [ ] When defined, display a background image in muskathlon detail page header
- [ ] Style payments success and aborted pages
- [ ] Validate input fields in payment modal
- [ ] Remove muskathlon list page (only admin access for now) ?

Improvements:
- [ ] Pretty format numbers (10'000 in place of 10000)
- [ ] Improve design for amount stats in participants detail page
- [ ] Do not copy css style into odoo, but use that from compassion website ?
- [ ] Clean code (remove some HTML tags and class copied from compassion website)

-> added to Odoo READMEs too.


ecino:
Can you put your comments "needs to be done" and "improvements" in the README of the different modules? This way, we will better remember it.

muskathlon/models/event_compassion.py
```diff
@@ -8,19 +8,63 @@
...
         string="Muskathlon event ID", size=128)
     muskathlon_registration_ids = fields.One2many(
         'muskathlon.registration', 'event_id', 'Muskathlon registrations')
 
+    public_description = fields.Text('Description publique')
+    picture_1 = fields.Binary('Picture 1')
+    picture_2 = fields.Binary('Picture 2')
+    video_url = fields.Char("Video URL")
+    participants_amount_objective = fields.Integer(
+        'Default raise objective by participant', default=10000, require=True)
```
ecino: the attribute is `required`. Please fix it everywhere you put `require`.

muskathlon/models/event_compassion.py
```diff
@@ -8,19 +8,63 @@
...
+                amount_raised += registration.amount_raised
+                amount_objective += registration.amount_objective
+
+            event.amount_raised = amount_raised
+            event.amount_objective = amount_objective
+            event.amount_raised_percents = int(
+                amount_raised * 100 / amount_objective)
+
+    @api.model
+    def getEventParticipants(self, event_id):
```
ecino: method names should be lowercase with underscores between words

muskathlon/models/event_compassion.py
```diff
@@ -46,3 +110,24 @@ class MuskathlonRegistration(models.Model):
...
          'Only one registration per participant/event is allowed!')
     ]
+
+    def _compute_amount_raised(self):
+        muskathlon_report = self.env['muskathlon.report']
+
+        for registration in self:
+            amount_raised = int(sum(
+                item.amount for item in muskathlon_report.search([]) if
+                item.user_id.id == registration.partner_id.id))
```
ecino: why not putting this condition in the search method? This would be more efficient.

payment_ogone_compassion/models/payment.py
```diff
@@ -0,0 +1,140 @@
...
+
+        res = ogone.ogone_form_generate_values(tx_data)
+
+        # filter field to send to Ogone
+        res = {k: v for k, v in res.iteritems() if k.isupper()}
+        del res['PARAMPLUS']
+        return ogone.ogone_get_form_action_url(), res
+
+    def validate_invoice_line(self, invoice_id):
+        invoice_line = self.env['account.invoice.line'].search([('id', '=',
```
ecino: Here the browse method does what you do with search. You can directly put the id and it fetches the record.

ecino (changes requested):

muskathlon/models/event_compassion.py
popod: Yeah, done.

muskathlon/models/event_compassion.py
popod: Oup's sorry..

payment_ogone_compassion/models/payment.py
popod: Done

muskathlon/models/event_compassion.py
popod: Yes, I've done it.

PR #428, opened 2018-04-25
CO-1643 : Add some tests for partner_compassion

samuel-fringeli:


partner_compassion/tests/test_partner_compassion.py
```diff
@@ -0,0 +1,86 @@
...
+            u'street': u'Impasse des Fr\xeanes 1',
+            u'zip': u'1669',
+            u'firstname': u'Samuel',
+            u'name': u'Fringeli Samuel',
+            u'city': u'Les Sciernes',
+            u'lastname': u'Fringeli',
+            u'ref': u'Fringeli',
+            u'church_id': self.church.id,
+            u'lang': u'en_US',
+            u'phone': u'+41 78 813 12 36'
```
ecino, line 36: Do you really want to put your information on Github? I think random data should be used...

ecino:
I guess that's a starting point... but coverage is not really changed with those small tests.

partner_compassion/tests/test_partner_compassion.py
vwvw: It happens even to the best ones ;)
https://github.com/CompassionCH/compassion-switzerland/blob/2845f2d19aa3be19680b9778dbca8efc07b372a1/wordpress_connector/models/contracts.py#L60-L71

ecino: :laughing: It's my secret plan to become famous!
@samuel-fringeli Are you ok with this? If good, I will merge.

PR #429, opened 2018-04-26
CO-1677 Paramter in URL

vwvw:
Tests are comming

### DIFF
--- a/muskathlon/models/event_compassion.py
+++ b/muskathlon/models/event_compassion.py
@@ -10,6 +10,7 @@
 ##############################################################################
 from odoo import models, fields, api
 from odoo.tools import config
+from odoo.exceptions import MissingError
 import re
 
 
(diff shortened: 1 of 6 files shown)


muskathlon/views/website_muskathlon_view.xml
```diff
@@ -171,7 +171,7 @@
 
                                         </div>
                                         <div style="margin-top: 10px;">
-                                            <a t-attf-href="https://compassion.ch/parrainer-un-enfant/?eventId={{event.id}}&amp;registrationId={{registration.id}}" target="_blank" class="button button-large">Sponsor a child</a>
+                                            <a t-attf-href="https://compassion.ch/parrainer-un-enfant/?eventId={{event.id}}&amp;registrationId={{registration.id}}&amp;participantId={{participant.id}}&amp;participantName={{participant.name}}" target="_blank" class="button button-large">Sponsor a child</a>
```
ecino: `https://compassion.ch/` can be replaced by the config `wordpress_host` present in odoo.conf file.
To achieve this I think the best way is to add a compute field in the registration object that will return the full URL.
This way we can redirect to test website on devel.

ecino (changes requested):

ecino:
Travis error is because of me and should be fixed with #432

muskathlon/models/event_compassion.py
```diff
@@ -125,6 +127,9 @@ def _compute_amount_raised(self):
             registration.amount_raised_percents = int(
                 amount_raised * 100 / registration.amount_objective)
 
+    def _compute_host(self):
+        config.get('wordpress_host')
```
ecino: the compute method should assign the value to the record, otherwise it won't work:
```
host = config.get('wordpress_host')
if not host:
    raise error
for registration in self:
    registration.host = host
```

ecino (changes requested):

ecino (approved):

PR #433, opened 2018-04-27
 CO-1676: modifications with muskathlons photos

samuel-fringeli:


ecino:
Your branch was not on par with website-muskathlon. You should rebase interactively and keep only your commit CO-1676. Remove all others.

ecino (approved):

PR #440, opened 2018-05-01
Fix for public access to the muskathlon website

vwvw:


muskathlon/models/event_compassion.py
```diff
@@ -78,6 +78,13 @@ class MuskathlonRegistration(models.Model):
         'res.partner', 'Muskathlon participant',
     )
 
+    # The 4 following fields avoid giving read access to the public on the
+    # res.partner participating in the muskathlon.
+    partner_id_id = fields.Integer(related="partner_id.id")
```
ecino: Please put all these fields the readonly attribute. We experienced several times security issues if related fields are not readonly, each time we write in the registration record, it will try to write as well in the related fields.

ecino (changes requested):

ecino (approved):

PR #445, opened 2018-05-02
Website muskathlon

vwvw:


muskathlon/models/event_compassion.py
```diff
@@ -89,8 +89,9 @@ class MuskathlonRegistration(models.Model):
         related="partner_id.preferred_name", readonly=True)
     partner_name = fields.Char(related="partner_id.name", readonly=True)
 
-    public_description = fields.Text('Public description')
-    ambassador_quote_public = fields.Text("Ambassador public quote")
+    public_description = fields.Text('Public description', translate=True)
```
ecino: Move it in crm.event.compassion please

muskathlon/models/event_compassion.py
```diff
@@ -89,8 +89,9 @@ class MuskathlonRegistration(models.Model):
...
+    ambassador_quote_public = fields.Text("Ambassador public quote",
```
ecino: Should be related to res.partner.ambassador quote (and I think in res.partner the field is already translatable, if not add the parameter)

ecino (changes requested):

ecino:
Please check the travis status, the tests are failing and there are pep8 errors.

PR #453, opened 2018-05-09
Current state of muskathlon registration form

vwvw:


ecino:
Cherry-picked the commit in my branch, thank you.

PR #460, opened 2018-05-17
CO-1695 [ADD] Module asterisk_autolog

eicher31:


hr_switzerland/__manifest__.py
```diff
@@ -39,8 +39,7 @@
         'views/hr_expense_custom.xml',
         'data/hr_config.xml'
     ],
-    'depends': [
-        'hr_expense', 'hr_attendance_extra_hours'],
+    'depends': ['hr_expense', 'hr_attendance_extra_hours', 'hr_attendance'],
```
ecino: `hr_attendance_extra_hours` already depends on `hr_attendance`, so adding the dependence is useless. However, you can add dependence to `asterisk_click2dial`

hr_switzerland/models/hr_attendance.py
```diff
@@ -0,0 +1,40 @@
...
+
+        return new_record
+
+    @api.multi
+    def write(self, vals):
+        """ Logout the agent related to the user during the checking """
+        user = self.employee_id.user_id
+        if (len(self) == 1) and ('check_out' in vals) and user.resource:
+            # todo: not enough check...
+            self.env['hr.asterisk_autolog'].agent_connexion(ast_user=user,
```
ecino: I think this will potentially disconnect users if they modifiy previous attendance. Why not using the field `attendance_state` in `hr.employee` to check if the user is logged in or not?

hr_switzerland/models/hr_autolog.py
```diff
@@ -0,0 +1,53 @@
...
+import logging
+
+from odoo import models, api, _
+from odoo.exceptions import UserError
+
+_logger = logging.getLogger(__name__)
+
+
+class asterisk_autolog(models.Model):
+    _name = 'hr.asterisk_autolog'
```
ecino: model names should only contain `.` instead of `_`

hr_switzerland/models/hr_autolog.py
```diff
@@ -0,0 +1,53 @@
...
+
+import logging
+
+from odoo import models, api, _
+from odoo.exceptions import UserError
+
+_logger = logging.getLogger(__name__)
+
+
+class asterisk_autolog(models.Model):
```
ecino: You can define it as `AbstractModel`, since there are no fields in this model. This way, it won't create a database table. But I would also say it could be possible as well to simply move the `agent_connexion` method in the `res.users` object, as a multi method. Then you could simply call `user.agent_connexion()`

ecino (changes requested):

hr_switzerland/models/hr_attendance.py
```diff
@@ -0,0 +1,43 @@
...
+            self.env['hr.asterisk.autolog'].agent_connexion(ast_user=user,
+                                                            connect=True)
+
+        return new_record
+
+    @api.multi
+    def write(self, vals):
+        """ Logout the agent related to the user during the checking """
+        super(HrAttendance, self).write(vals)
+        user = self.employee_id.user_id
```
ecino: Here it will fail if self contains multiple record (although you test it just in the next line). I would suggest to add a for loop to respect the api.multi of the write method. `for attendance in self:`

hr_switzerland/models/hr_attendance.py
```diff
@@ -0,0 +1,43 @@
...
+        if (len(self) == 1) and ('check_out' in vals) and user.resource:
+            # todo: not enough check...
+            if self.employee_id.attendance_state == 'checked_out':
+                user.asterisk_connect(False)
+                # self.env['hr.asterisk.autolog'].agent_connexion(ast_user=user,
+                #                                                 connect=False)
+        return
```
ecino: Write method should return `True`

hr_switzerland/models/res_users.py
```diff
@@ -0,0 +1,52 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class ResUser(models.Model):
+    _inherit = 'res.users'
+
+    @api.model
+    def asterisk_connect(self, log=True):
+        for ast_user in self:
```
ecino: Maybe filter only users that have `resource` field? `for ast_user in self.filtered('resource'):`

hr_switzerland/models/res_users.py
```diff
@@ -0,0 +1,52 @@
...
+                    timeout=unicode(ast_server.wait_time * 1000),
+                    caller_id=ast_user.internal_number,
+                    account=ast_user.cdraccount)
+            except Exception, e:
+                _logger.error(
+                    "Error in the Originate request to Asterisk server %s",
+                    ast_server.ip_address)
+                _logger.error(
+                    "Here are the details of the error: '%s'", unicode(e))
+                raise UserError(
```
ecino: I'm not sure that we should raise the error? It means the user won't be able to check in or check out in case of error. I think it's okay if the asterisk logon/logoff fails. What do you think? Or maybe here we could use a small toast message if we inherit from this module that I looked for some time : [OCA/web_notify](https://github.com/OCA/web/tree/10.0/web_notify)

ecino (changes requested):

ecino (commented):
Needs changes according to the review in compassion-modules. I think the web_notify will work in production, we should test in devel. `This module is based on the Instant Messaging Bus. To work properly, the server must be launched in gevent mode.` But if we cannot use the gevent mode maybe we should remove this, we have to check.

ecino (approved):

PR #473, opened 2018-05-23
CO-1741: picture_1 in ambassador view if picture_2 doesn't exist

samuel-fringeli:


muskathlon/views/website/muskathlon_participant_details.xml
```diff
@@ -26,22 +26,29 @@
...
+                                        <div class="col-lg-1"/>
                                         <div class="col-lg-8 col-sm-11">
                                             <div class="row-fluid">
                                                 <div class="col-sm-6" style="min-width: 330px;">
                                                     <div class="image-and-buttons" style="padding-right: 30px;">
-                                                        <div class="image"
-                                                             style="display: inline-block; max-height: 310px;"
-                                                             t-field="registration.ambassador_picture_2"
-                                                             t-field-options='{"widget": "image"}'/>
+                                                        <t t-if="not registration.ambassador_picture_2">
```
ecino, line 34: Does it work to do instead
```xml
<t t-set="ambassador_image" value="registration.ambassador_picture_2 or registration.ambassador_picture_1"/>
<div class="image"
style="display: inline-block; max-height: 310px;"
 t-field="ambassador_image"
 t-field-options='{"widget": "image"}'/>
```
This would be more clean if it works.

samuel-fringeli: Unfortunately, it doesn't work : 
```
Error when compiling AST
AssertionError: t-field must have at least a dot like 'record.field_name'
```

ecino: Fine then.

ecino (approved):

PR #482, opened 2018-06-01
CO-1681 [update] Wordpress action  valid privacy statement

eicher31:


wordpress_connector/models/import_letters_history.py
```diff
@@ -177,18 +177,8 @@ def import_web_letter(self, child_code, sponsor_ref, name, email,
...
-                contract.agreement_date = fields.Date.today()
-            else:
-                statement = self.env[
-                    'compassion.privacy.statement'].get_current()
-                self.env['privacy.statement.agreement'].create(
-                    {'partner_id': sponsor_id,
-                     'agreement_date': fields.Date.today(),
-                     'privacy_statement_id': statement.id})
+            model_sponsor[:1].id.set_privacy_statement(
+                origin='Website partner creation')
```
ecino: The origin would be website lettrer, no?

wordpress_connector/models/contracts.py
```diff
@@ -234,6 +234,8 @@ def create_sponsor_from_web(self):
         })
         self.child_id.sponsor_id = partner
 
+        partner.set_privacy_statement(origin='Website partner creation')
```
ecino: There is the case of a new sponsorship that is missing. Here it will do it only if it's a new partner, but what about when an existing partner makes a new sponsorship?

ecino (changes requested):

ecino:
Please check the travis error. Is it because the changes from compassion-modules are not yet merged into 10.0?

ecino (approved):

PR #483, opened 2018-06-04
CO-1734: changes in muskathlon website design

samuel-fringeli:


muskathlon/models/muskathlon_registration.py
```diff
@@ -50,6 +50,15 @@ class MuskathlonRegistration(models.Model):
         related='ambassador_details_id.description', readonly=True)
     ambassador_quote = fields.Text(
         related='ambassador_details_id.quote', readonly=True)
+    ambassador_thank_you_quote = fields.Html(
+        related='ambassador_details_id.thank_you_quote', readonly=True,
+        default='Thank you for your donation !')
```
ecino: This related field is not useful, the original field is only used in donation letters (computed field with quote text + ambassador picture). You can remove it.

ecino (changes requested):

ecino (approved):

PR #485, opened 2018-06-05
muskathlon fixes

samuel-fringeli:


ecino (changes requested):
It would be better to have all muskathlon styles in a css file inside the muskathlon module, instead of putting it in website_compassion.
Website_compassion should only contain styles that are usable in all website.

muskathlon/__manifest__.py
```diff
@@ -50,6 +50,7 @@
         'views/ambassador_details_view.xml',
         'views/notification_settings_view.xml',
         'views/payment_transaction_view.xml',
+        'views/assets.xml',
```
ecino: No, you should include this in `templates/assets` which is already existing.

ecino (changes requested):

ecino (approved):

PR #530, opened 2018-06-26
CO-1820: send one welcome letter per day

samuel-fringeli:


partner_communication_switzerland/models/contracts.py
```diff
@@ -246,6 +246,17 @@ def send_daily_communication(self):
...
 
+        # send welcome activations letters
+        welcome = self.env.ref(
+            'partner_communication_switzerland.welcome_activation')
+        to_send = self.env['recurring.contract'].search([
+            ('activation_date', '>=', (datetime.today() - timedelta(days=1))),
+            ('child_id', '!=', False)
+        ])
+        if to_send:
+            to_send.send_communication(welcome, both=True)
```
ecino: Here you should do
```
to_send.send_communication(welcome, both=True).send()
```
To force sending the e-mail rightaway.

ecino (changes requested):

ecino (approved):

PR #544, opened 2018-07-11
CO-1700 Move field invoice_type

eicher31:
in compassion_module/sponsorship_compassion

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -26,5 +26,5 @@ survey
 bank-payment
 web https://github.com/daramousk/web 10.0-MIG-CKEditor
 l10n-switzerland https://github.com/CompassionCH/l10n-switzerland 10.0-pain-sepa
-compassion-accounting https://github.com/CompassionCH/compassion-accounting 10.0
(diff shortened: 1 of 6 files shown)


ecino (changes requested):
One import that became useless:
sponsorship_switzerland/models/account_invoice.py:16:1: F401 'odoo.addons.sponsorship_compassion.models.product.FUND_CATEGORY' imported but unused

You should as well move the field from the view account_invoice_view.xml,.

sponsorship_switzerland/__manifest__.py
```diff
@@ -38,6 +38,7 @@
     'depends': [
         'crm_compassion',
         'sponsorship_tracking',
+        'sponsorship_compassion',
```
ecino: `sponsorship_tracking` already depends on `sponsorship_compassion`. This addition is not needed.

ecino (changes requested):
Change dependencie to branch devel on file `oca_dependencies.txt`

ecino (approved):

PR #548, opened 2018-07-16
CO-1901 fix attempt for welcome active letters for Demaurex

quentingigon:


partner_communication_switzerland/models/contracts.py
```diff
@@ -223,7 +223,8 @@ def send_daily_communication(self):
             '|', ('correspondent_id.email', '!=', False),
             ('partner_id.email', '!=', False),
             ('state', '=', 'active'),
-            ('type', 'like', 'S')
+            ('type', 'like', 'S'),
+            ('partner_id', '!=0', 1502623) # if partner is not Demaurex
```
ecino: 1. You have a typo: '!=0'
2. Condition field is 'partner_id.ref'
3. 1502623 should be a quoted string
4. pylint issue: you should put 2 spaces before inline comment

ecino (changes requested):

partner_communication_switzerland/models/contracts.py
quentingigon: That was a really bad commit.

ecino (approved):

PR #563, opened 2018-07-25
CO-1892 Update the CRM stats

nfluckiger:
Add link between different stats object


mass_mailing_switzerland/__manifest__.py
```diff
@@ -39,11 +39,15 @@
...
         'views/mail_template_view.xml',
         'views/utm_view.xml',
+        'views/contract_origin_view.xml',
         'data/refresh_sendgrid_tracking_cron.xml',
     ],
     'depends': [
         'mail_sendgrid_mass_mailing',
-        'partner_communication_switzerland'
+        'partner_communication_switzerland',
+        'sponsorship_compassion',
```
ecino: `partner_communication_switzerland` already depends on the three modules you added, so it's not needed to add these dependencies.

mass_mailing_switzerland/models/correspondence.py
```diff
@@ -0,0 +1,16 @@
...
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+from odoo import models, fields
+
+
+class Correspondence(models.Model):
+    _inherit = 'correspondence'
+    campaign_id = fields.Many2one('umt.campaign')
```
ecino: `utm.campaign`

mass_mailing_switzerland/models/crm_event_compassion.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class EventCompassion(models.Model):
+
+    _inherit = 'crm.event.compassion'
+
+    @api.model
+    def create(self, vals):
+        event = super(EventCompassion, self).create(vals)
+        for NewEvent in self:
```
ecino: `variable_name` should be lowercase with underscores (code convention by odoo). The loop is not needed because create can only create one record. You can simply use `event` instead of new_event.

mass_mailing_switzerland/models/crm_event_compassion.py
```diff
@@ -0,0 +1,32 @@
...
+            if NewEvent.campaign_id is not None:
```
ecino: `if event.campaign_id` is enough for the test. Relational fields are equal to False in case they are not set.

mass_mailing_switzerland/models/crm_event_compassion.py
```diff
@@ -0,0 +1,32 @@
...
+                NewEvent.analytic_id.campaign_id = NewEvent.campaign_id
+        return event
+
+    @api.multi
+    def write(self, vals):
+        event = super(EventCompassion, self).write(vals)
+        for NewEvent in self:
```
ecino: `new_event`

mass_mailing_switzerland/models/crm_event_compassion.py
```diff
@@ -0,0 +1,32 @@
...
+    def create(self, vals):
+        event = super(EventCompassion, self).create(vals)
+        for NewEvent in self:
+            if NewEvent.campaign_id is not None:
+                NewEvent.analytic_id.campaign_id = NewEvent.campaign_id
+        return event
+
+    @api.multi
+    def write(self, vals):
+        event = super(EventCompassion, self).write(vals)
```
ecino: Rename this as the write method doesn't return a record. It returns True, so you can name it `res` for instance.

mass_mailing_switzerland/models/crm_event_compassion.py
```diff
@@ -0,0 +1,32 @@
...
+        for NewEvent in self:
+            if NewEvent.campaign_id is not None:
```
ecino: `if event.campaign_id:`

mass_mailing_switzerland/models/utm.py
```diff
@@ -58,17 +58,17 @@ class UtmObjects(models.AbstractModel):
         readonly=True
     )
     invoice_ids = fields.One2many(
```
ecino: This should be renamed to `invoice_line_ids`. Beware to rename usages as well.

mass_mailing_switzerland/models/mass_mailing_campaign.py
```diff
@@ -27,7 +27,7 @@ class MassMailingCampaign(models.Model):
         'correspondence', related='campaign_id.correspondence_ids'
     )
     invoice_ids = fields.One2many(
```
ecino: Should be renamed to `invoice_line_ids`. Beware to rename usages as well.

mass_mailing_switzerland/models/mass_mailing_campaign.py
```diff
@@ -27,7 +27,7 @@ class MassMailingCampaign(models.Model):
...
-        'account.invoice', related='campaign_id.invoice_ids'
+        'account.invoice.line', related='campaign_id.invoice_ids'
```
ecino: related field would become `campaign_id.invoice_line_ids` when the field in utm will be renamed as well.

mass_mailing_switzerland/models/recurring_contract.py
```diff
@@ -0,0 +1,48 @@
...
+#
+##############################################################################
+from odoo import models, fields, api
+
+
+class RecurringContract(models.Model):
+
+    _inherit = 'recurring.contract'
+
+    campaign_id = fields.Many2one('utm.campaign')
```
ecino: This is not needed because `recurring.contract` inherits from `utm.mixin` and this adds all relation to utm objects.

mass_mailing_switzerland/models/account_invoice_line.py
```diff
@@ -0,0 +1,18 @@
...
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+from odoo import models, fields
+
+
+class AccountInvoiceLine(models.Model):
+    _inherit = 'account.invoice.line'
+    campaign_id = fields.Many2one('umt.campaign')
```
ecino: Instead of adding the 3 relations you can simply inherit from `utm.mixin` object. This is exactly made for inheriting those 3 fields. You can do this like this:
```
_inherit = ['account.invoice.line', 'utm.mixin']
_name = 'account.invoice.line'
```
And you can remove the fields.

mass_mailing_switzerland/models/recurring_contract.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @api.model
+    def create(self, vals):
+        origin = self.env['recurring.contract.origin'].search([('id', '=', vals['origin_id'])])
```
ecino: This will fail if `origin_id` is not set at creation. You should first test `if 'origin_id' in vals:`

mass_mailing_switzerland/models/recurring_contract.py
```diff
@@ -0,0 +1,48 @@
...
+        if origin.type == 'event':
+            vals['campaign_id'] = origin.event_id.campaign_id.id
+        if origin.type == 'marketing':
+            analytic = self.env['account.analytic.account'].search([('id', '=', origin.analytic_id.id)])
```
ecino: Here you can simply do `origin.analytic_id.campaign_id = origin.campaign_id`

mass_mailing_switzerland/models/recurring_contract.py
```diff
@@ -0,0 +1,48 @@
...
+        return super(RecurringContract, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        if 'origin_id' in vals:
+            origin = self.env['recurring.contract.origin'].search([('id', '=', vals['origin_id'])])
+            if origin.type == 'event':
+                vals['campaign_id'] = origin.event_id.campaign_id.id
+            if origin.type == 'marketing':
+                analytic = self.env['account.analytic.account'].search([('id', '=', origin.analytic_id.id)])
```
ecino: Here you can simply do `origin.analytic_id.campaign_id = origin.campaign_id`

mass_mailing_switzerland/models/recurring_contract.py
```diff
@@ -0,0 +1,48 @@
...
+                analytic.write({'campaign_id': origin.campaign_id.id})
+
+        return super(RecurringContract, self).write(vals)
+
+
+class RecurringContractOrigin(models.Model):
+
+    _inherit = 'recurring.contract.origin'
+
+    campaign_id = fields.Many2one('utm.campaign')
```
ecino: Add string parameter `Many2one('utm.campaign', 'Campaign')` to display `Campaign` in the view instead of `Campaign id`

mass_mailing_switzerland/models/utm.py
```diff
@@ -80,11 +80,13 @@ def _compute_letters_count(self):
         for utm in self:
             utm.letters_count = len(utm.correspondence_ids)
 
-    @api.depends('invoice_ids', 'invoice_ids.amount_total')
+    @api.depends('invoice_ids')
```
ecino: depends is no more needed since the field is not stored

mass_mailing_switzerland/views/contract_origin_view.xml
```diff
@@ -0,0 +1,23 @@
...
+                                                          'required': [('type', 'in', ('marketing', 'event'))]}"/>
+            </field>
+        </field>
+    </record>
+  <record id="view_account_analytic_account_form" model="ir.ui.view">
+    <field name="model">account.analytic.account</field>
+    <field name="inherit_id" ref="analytic.view_account_analytic_account_form"/>
+    <field name="arch" type="xml">
+        <field name="partner_id" position="after">
+            <field name="campaign_id" string="Campaign"/>
```
ecino: You don't need the string parameter if you define it in the model.

mass_mailing_switzerland/views/contract_origin_view.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record id="view_contract_origin_form_mass_mailing" model="ir.ui.view">
+        <field name="model">recurring.contract.origin</field>
+        <field name="inherit_id" ref="contract_compassion.view_contract_origin_form"/>
+        <field name="arch" type="xml">
+            <field name="analytic_id" position="after">
+                <field name="campaign_id" string="Campaign"  attrs="{'invisible': [('type', 'not in', ('marketing','partner', 'event'))],
```
ecino: You don't need the string parameter if you define it in the model.

mass_mailing_switzerland/models/utm.py
```diff
@@ -196,6 +207,26 @@ def _compute_click_count(self):
             campaign.click_count = sum(self.env['link.tracker'].search([
                 ('campaign_id', '=', campaign.id)]).mapped('count'))
 
+    def _compute_contracts(self):
+        for campaign in self:
+            campaign.contract_ids = self.env['recurring.contract'].\
+                search(['|', ('campaign_id.id', '=', campaign.id),
```
ecino: .id is not needed in the left part (it's like a query in sql)

mass_mailing_switzerland/models/utm.py
```diff
@@ -196,6 +207,26 @@ def _compute_click_count(self):
...
+                        ('origin_id.event_id.campaign_id.id', '=', campaign.id)
```
ecino: .id is not needed in the left part

mass_mailing_switzerland/models/utm.py
```diff
@@ -196,6 +207,26 @@ def _compute_click_count(self):
...
+                        ])
+
+    def _compute_correspondence(self):
+        for campaign in self:
+            campaign.correspondence_ids = self.env['correspondence'].search([
+                '|', ('sponsorship_id.campaign_id.id', '=', campaign.id),
```
ecino: .id is not needed in the left part

mass_mailing_switzerland/models/utm.py
```diff
@@ -196,6 +207,26 @@ def _compute_click_count(self):
...
+                ('sponsorship_id.origin_id.event_id.campaign_id.id', '=',
```
ecino: .id is not needed in the left part

mass_mailing_switzerland/models/account_invoice_line.py
ecino: I don't see anywhere where you set those values in the invoice.line ? I'm not sure those fields are actually needed?

mass_mailing_switzerland/models/utm.py
```diff
@@ -196,6 +207,26 @@ def _compute_click_count(self):
...
+                 campaign.id)])
+
+    def _compute_invoice(self):
+        for campaign in self:
+            campaign.invoice_ids = self.env['account.invoice.line'].search([
+                '|', ('invoice_id.campaign_id', '=', campaign.id),
+                ('event_id.campaign_id', '=', campaign.id)])
```
ecino: Missing or condition `('account_analytic_id.campaign_id', '=', campaign.id)`

ecino (changes requested):
Some changes to do, please verify as well travis status check.

mass_mailing_switzerland/models/recurring_contract.py
```diff
@@ -0,0 +1,48 @@
...
+class RecurringContract(models.Model):
+
+    _inherit = 'recurring.contract'
+
+    campaign_id = fields.Many2one('utm.campaign')
+
+    @api.model
+    def create(self, vals):
+        origin = self.env['recurring.contract.origin'].search([('id', '=', vals['origin_id'])])
+        if origin.type == 'event':
```
ecino: `if origin.type == 'event' and 'campaign_id' not in vals:` to avoid replacing the campaign if it is set at creation.

mass_mailing_switzerland/models/recurring_contract.py
```diff
@@ -0,0 +1,48 @@
...
+            analytic = self.env['account.analytic.account'].search([('id', '=', origin.analytic_id.id)])
+            analytic.write({'campaign_id': origin.campaign_id.id})
+
+        return super(RecurringContract, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        if 'origin_id' in vals:
+            origin = self.env['recurring.contract.origin'].search([('id', '=', vals['origin_id'])])
+            if origin.type == 'event':
```
ecino: `if origin.type == 'event' and not self.mapped('campaign_id')` to avoid replacing a campaign that is already set.

ecino (changes requested):

ecino (approved):

PR #572, opened 2018-08-06
Devel to production - new UTM stats + SMS communications 

ecino:


report_compassion/report/partner_communication.xml
```diff
@@ -52,15 +52,15 @@
                 font-family: "tetriaLight";
                 font-size: 10pt;
                 <t t-if="success_story">
-                    margin-left: 6mm;
-                    margin-right: 6mm;
+                    margin-left: 8mm;
```
ecino: With this margin change, the communication titles are on two lines. We can see if we can improve this but be careful that the title doesn't show in the address of the envelope.

PR #587, opened 2018-08-14
Co 1897 Partner can have more than one email and one birthdate

nfluckiger:


ecino:
You created a branch from your other branch with editor changes. When you create a new branch, make sure you start from compassion/devel state and not one of your local branches, otherwise you take the commits inside your new branch.

Please rebase onto compassion/devel and pick only the relevant commits.

ecino (approved):
I will merge after we put in production the current state of devel. Then we will test it. Thanks!

PR #594, opened 2018-08-16
CO-1995 Changed welcome active letter generation

quentingigon:


partner_communication_switzerland/migrations/10.0.2.4.6/post-migration.py
```diff
@@ -0,0 +1,27 @@
...
+def migrate(env, version):
+    if not version:
+        return
+
+    sponsorships = env['recurring.contract'].search([
+        ('activation_date', '!=', None)
+    ])
+    for sponsorship in sponsorships:
+        # if activation date is more than 24h old
+        if sponsorship.activation_date <= fields.Datetime.to_string(
```
ecino: Here you compare strings. It's safer to compare date objects to determine which is older. You can use `fields.Datetime.from_string(sponsorship.activation_date)`

partner_communication_switzerland/models/contracts.py
```diff
@@ -417,7 +427,8 @@ def contract_waiting(self):
                       s.correspondent_id.email and s.partner_id.ref !=
                       '1502623')
         welcome.write({
-            'sds_state': 'waiting_welcome'
+            'sds_state': 'waiting_welcome',
+            'start_date': datetime.today()
```
ecino: This should be done in module `recurring_contract` instead.

sponsorship_switzerland/models/contracts.py
```diff
@@ -213,7 +213,10 @@ def contract_active(self):
 
     @api.multi
     def contract_waiting_mandate(self):
-        self.write({'state': 'mandate'})
+        self.write({
+            'state': 'mandate',
+            'start_date': datetime.today()
```
ecino: This should be done in module `sponsorship_switzerland` instead.

partner_communication_switzerland/migrations/10.0.2.4.6/post-migration.py
```diff
@@ -0,0 +1,27 @@
...
+from datetime import datetime, timedelta
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    if not version:
+        return
+
+    sponsorships = env['recurring.contract'].search([
```
ecino: You could replace this whole script by a SQL query which will be much faster :

1. Remove openupgrade.migrate decorator
2. `env` function parameter becomes `cr`
3.

```
cr.execute("""
update recurring_contract
set welcome_active_letter_sent=true
where activation_date <= %s
""", fields.Datetime.to_string(datetime.today() - timedelta(days=1)))
```

ecino (changes requested):

sponsorship_switzerland/models/contracts.py
quentingigon: We are in module `sponsorship_switzerland`, no ?

quentingigon:
So for the start_date, now I set it up in compassion-accounting/recurring_contract/recurring_contract in the write method

ecino (approved):

PR #608, opened 2018-08-30
fix general translation linting errors

nbornand:


partner_compassion/models/partner_bank_compassion.py
```diff
@@ -26,7 +26,7 @@ def create(self, data):
 
         part = result.partner_id
         part.message_post(_("<b>Account number: </b>" + result.acc_number),
-                          _("New account created"), 'comment')
+                          _("New account created"), _('comment'))
```
ecino, line 29: 'comment' shouldn't be translated, it's a type of message

partner_compassion/models/partner_bank_compassion.py
```diff
@@ -38,7 +38,7 @@ def unlink(self):
             part = account.partner_id
             part.message_post(_("<b>Account number: </b>" +
                                 account.acc_number),
-                              _("Account deleted"), 'comment')
+                              _("Account deleted"), _('comment'))
```
ecino, line 41: 'comment' shouldn't be translated

ecino (changes requested):

PR #618, opened 2018-09-05
CO-2024 Translators migration to advocate details

maxime-beck:


partner_compassion/models/ambassador_details.py
```diff
@@ -14,6 +14,7 @@
 
 from odoo import api, models, fields, _
 from odoo.tools import file_open
+from odoo.addons.sbc_switzerland.models import translate_connector
```
ecino: No, you should rather inherit from `advocate.details` in the module `sbc_switzerland` and there inherit the `write` method

ecino (changes requested):

ecino (approved):

PR #639, opened 2018-10-02
CO-2052 name splitted in muskathlon form

quentingigon:


ecino (changes requested):
There is also the muskathlon_donation_form that should be updated!

ecino (approved):

PR #642, opened 2018-10-02
C-2060 BVR ref is created at initialization

quentingigon:


sms_939/models/recurring_contract_group.py
```diff
@@ -0,0 +1,29 @@
...
+from odoo import api, models, _
+from odoo.exceptions import UserError
+
+
+class ContractGroup(models.Model):
+    _inherit = 'recurring.contract.group'
+
+    @api.multi
+    @api.onchange('payment_mode_id')
+    def onchange_payment_mode_id(self):
```
ecino: This method already exists here : https://github.com/CompassionCH/compassion-switzerland/blob/10.0/sponsorship_switzerland/models/contract_group.py#L149

You can delete this file.

sms_939/models/recurring_contract.py
```diff
@@ -8,7 +8,8 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
-from odoo import api, models
+from odoo import api, models, _
+from odoo.exceptions import UserError
```
ecino: this change is not needed

sms_939/models/__init__.py
```diff
@@ -13,4 +13,5 @@
 from . import sms_notification
 from . import compassion_child
 from . import recurring_contract
```
ecino: In the model recurring_contract you should inherit the method `associate_group` and call the method `on_change_payment_mode` from there.

ecino (changes requested):

ecino:
Please rebase onto compassion/devel and fix conflicts

quentingigon:
I dont know if the problem comes from my modification but when testing I found that when we use an existant partner for an SMS request, the first time we complete the second form fails (but with no error message) and no children is assigned to the partner. If we redo the form, everything is working as intended. This does not happen with a partner just created

ecino (approved):

PR #654, opened 2018-10-09
CO-2095 Change sms provider

quentingigon:


quentingigon:
I couldn't test with my phone to see if the correct account is selected (small one) for an automated response but it should be the case

partner_communication_switzerland/wizards/generate_communication_wizard.py
```diff
@@ -53,6 +53,10 @@ class GenerateCommunicationWizard(models.TransientModel):
     sms_cost_estimation = fields.Float(compute='_compute_sms_cost_estimation')
     currency_id = fields.Many2one('res.currency', compute='_compute_currency')
     campaign_id = fields.Many2one('utm.campaign', 'Campaign')
+    sms_provider = fields.Many2one('sms.provider', 'SMS Provider',
```
ecino: You added the field in the view but the value should be passed  to the sms wizard when generating the communication!

partner_communication_switzerland/wizards/generate_communication_wizard.py
```diff
@@ -53,6 +53,10 @@ class GenerateCommunicationWizard(models.TransientModel):
...
+                                   default=lambda self: self.env
+                                   ['sms.provider'].search([('id', '=', 1)]),
```
ecino: never hardcode ids. Here you can do `search([], limit=1)`

sms_939/models/sms_notification.py
```diff
@@ -124,7 +124,8 @@ def run_service(self):
         if 'test' not in sms_text:
             self.env['sms.sender.wizard'].create({
                 'text': sms_answer
-            }).send_sms(mobile=sms_receipient)
+                # large account id is 1, small account id is 2
+            }).send_sms(mobile=sms_receipient, provider_account_id=2)
```
ecino: never hardcode ids. the provider should be taken into to the linked sms_hook (self.hook_id)

sms_939/wizards/sms_sender_wizard.py
```diff
@@ -52,14 +55,19 @@ def _compute_partner(self):
                 raise UserError(_("No valid partner"))
 
     @api.multi
-    def send_sms(self, mobile):
+    def send_sms(self, mobile, provider_account_id=None):
```
ecino: That's not the good way of doing it. The method should not accept parameter and will only use self.sms_provider. It's up to the caller to set properly the field before calling the method.

sms_939/wizards/sms_sender_wizard.py
```diff
@@ -52,14 +55,19 @@ def _compute_partner(self):
...
         self.ensure_one()
         if not mobile:
             return False
 
+        if provider_account_id:
```
ecino: Remove this portion

sms_939/models/sms_notification.py
ecino: In the create you want to set the field sms_provder_id and not in the send_sms method.

ecino (changes requested):
Please update the tests as well.

partner_communication_switzerland/models/partner_communication.py
```diff
@@ -42,6 +42,7 @@ class PartnerCommunication(models.Model):
     currency_id = fields.Many2one('res.currency', compute='_compute_currency')
     utm_campaign_id = fields.Many2one('utm.campaign')
     sms_cost = fields.Float()
+    sms_provider = fields.Many2one('sms.provider', 'SMS Provider')
```
ecino: Many2one fields should be suffixed by `_id` by convention. Please rename all your fields `sms_provider` to `sms_provider_id`

ecino: Add the default value like you did for the wizard

ecino: Add the field in the communication_job view

partner_communication_switzerland/models/partner_communication.py
```diff
@@ -433,7 +434,8 @@ def send_by_sms(self):
             sms_wizard = self.env['sms.sender.wizard'].with_context(
                 partner_id=job.partner_id.id).create({
                     'subject': job.subject,
-                    'text': sms_text
+                    'text': sms_text,
+                    'sms_provider': job.sms_provider
```
ecino: should be `job.sms_provider_id.id` (integer value is passed for the create method)

partner_communication_switzerland/wizards/generate_communication_wizard.py
```diff
@@ -150,6 +151,7 @@ def generate_communications(self, async_mode=True):
                     'send_mode': self.send_mode,
                     'report_id': self.report_id.id or
                     self.model_id.report_id.id,
+                    'sms_provider': self.sms_provider
```
ecino: you should pass the id

sms_939/models/sms_notification.py
```diff
@@ -123,9 +123,11 @@ def run_service(self):
                     })
         if 'test' not in sms_text:
             self.env['sms.sender.wizard'].create({
-                'text': sms_answer
-                # large account id is 1, small account id is 2
-            }).send_sms(mobile=sms_receipient, provider_account_id=2)
+                'text': sms_answer,
+                'sms_provider': self.env['sms.provider'].search([
+                    ('name', 'ilike', 'small account')
```
ecino: Not a good idea either, because we have to make sur that we name our provider  "small account", or this will fail to find the good one. I think the best way to differentiate the two providers is to add a xml data file and create the two records (and fill the fields "name", "server", "port", "endpoint" but put dummy values in fields "username" and "password").
Then here and other places where you need to reference the correct provider, you can use the function `self.env.ref('sms_939.provider_xml_id')` fo fetch the correct provider.

ecino (changes requested):

quentingigon:
I will add the tests next week

ecino:
```
The model sms.provider has no access rules, consider adding one. E.g. access_sms_provider,access_sms_provider,model_sms_provider,base.group_user,1,0,0,0
```

partner_communication_switzerland/models/partner_communication.py
```diff
@@ -42,6 +42,10 @@ class PartnerCommunication(models.Model):
     currency_id = fields.Many2one('res.currency', compute='_compute_currency')
     utm_campaign_id = fields.Many2one('utm.campaign')
     sms_cost = fields.Float()
+    sms_provider_id = fields.Many2one('sms.provider', 'SMS Provider',
+                                      default=lambda self: self.
```
ecino, line 46:
```suggestion
                                      default=lambda self: self.
                                      env.ref('sms_939.large_account_id').id,
```

partner_communication_switzerland/models/partner_communication.py
```diff
@@ -42,6 +42,10 @@ class PartnerCommunication(models.Model):
...
+                                      env['sms.provider'].search([], limit=1),
```
ecino:
```suggestion
```

partner_communication_switzerland/wizards/generate_communication_wizard.py
```diff
@@ -53,6 +53,10 @@ class GenerateCommunicationWizard(models.TransientModel):
     sms_cost_estimation = fields.Float(compute='_compute_sms_cost_estimation')
     currency_id = fields.Many2one('res.currency', compute='_compute_currency')
     campaign_id = fields.Many2one('utm.campaign', 'Campaign')
+    sms_provider_id = fields.Many2one('sms.provider', 'SMS Provider',
+                                      default=lambda self: self.
+                                      env['sms.provider'].search([], limit=1),
```
ecino:
```suggestion
                                      env.ref('sms_939.large_account_id').id,
```

sms_939/wizards/sms_sender_wizard.py
```diff
@@ -43,6 +42,10 @@ class SmsSender(models.TransientModel):
     partner_id = fields.Many2one(
         'res.partner', 'Partner', compute='_compute_partner')
     sms_request_id = fields.Many2one(comodel_name='sms.child.request')
+    sms_provider_id = fields.Many2one('sms.provider', "SMS Provider",
+                                      default=lambda self: self.
+                                      env['sms.provider'].search([], limit=1),
```
ecino:
```suggestion
                                      env.ref('sms_939.large_account_id').id,
```

ecino (changes requested):

ecino (approved):

PR #655, opened 2018-10-09
CO-2099 Add option to active the Xivo agent or not

eicher31:


hr_switzerland/models/res_users.py
```diff
@@ -9,41 +9,46 @@
...
-                    account=ast_user.cdraccount)
-                message = 'Your Xivo agent is now {}.'.format(
-                    'connected' if log else 'disconnected')
-                ast_user.notify_info(message)
-                ast_manager.Logoff()
-            except Exception, e:
-                message = "Impossible to connect your Xivo agent\n"
-                message += unicode(e)
-                ast_user.notify_info(message)
+            if ast_user.connect_agent:
```
ecino: Instead of adding this `if` clause in the loop, you can change the loop to `for ast_user in self.filtered('connect_agent')`

ecino (commented):
The tests will succeed when your changes will be merged in branch 10.0 in compassion-modules.

PR #666, opened 2018-10-17
Co 2086 Now the language work in the linked partner & Co 2038 Do not set the Event on a thank you letter for a gift

nfluckiger:


partner_communication_switzerland/models/account_invoice_line.py
```diff
@@ -75,11 +75,16 @@ def generate_thank_you(self):
         event = invoice_lines.mapped('event_id')
         ambassadors = invoice_lines.mapped('user_id')
 
+        event_id = event.id
+
+        if invoice_lines.mapped('gift_id'):
+            event_id = None
```
ecino: event_id = False, is better for Odoo Many2one values

partner_communication_switzerland/models/account_invoice_line.py
```diff
@@ -75,11 +75,16 @@ def generate_thank_you(self):
         event = invoice_lines.mapped('event_id')
         ambassadors = invoice_lines.mapped('user_id')
 
+        event_id = event.id
+
+        if invoice_lines.mapped('gift_id'):
```
ecino: Here you could look for contract instead of only gift, meaning that any donation related to a sponsorship is not considered as an event donation:
`if invoice_lines.mapped('contract_id'):`

ecino (changes requested):

ecino (approved):

PR #683, opened 2018-10-31
CO-2090 Search on linked partner work now with archived partner

eicher31:


ecino (changes requested):
I am missing the same when

- sponsorship is made from wordpress
- sponsorship is made from muskathlon
- donation is made for muskathlon

For the last two it should be done in module compassion-modules/cms_form_compassion in the match_partner_form.py where we find the matching partner.

ecino (approved):

PR #687, opened 2018-11-13
CO-2148 Muskathlon objective

quentingigon:


muskathlon/models/event_compassion.py
```diff
@@ -54,9 +54,9 @@ def _compute_amount_raised(self):
                 amount_objective += registration.amount_objective
 
             event.amount_raised = amount_raised
-            event.amount_objective = amount_objective
+            event.amount_objective = amount_objective or 10000
```
ecino: Not a good fix, the problem is that registration.amount_objective is 0 by default and it shouldn't when a participant register. It should take the default value that is set in the event.event object.

ecino (changes requested):

ecino:
Fixed with #690

PR #697, opened 2018-11-21
CO-2145 Translate addresses for IT and DE

nfluckiger:
Update two function to be able to translate the address for Bern office in IT and DE

### DIFF
--- a/report_compassion/models/res_partner.py
+++ b/report_compassion/models/res_partner.py
@@ -8,9 +8,11 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
-from datetime import date
+from locale import setlocale
 
-from odoo import api, models, fields
+from datetime import date, datetime
+
+from odoo import api, models, fields, _
 
 
 class ResPartner(models.Model):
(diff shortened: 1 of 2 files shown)


report_compassion/models/res_partner.py
```diff
@@ -43,3 +45,23 @@ def get_receipt(self, year):
...
+            'fr_CH': u'le %d %B %Y',
+            'fr': u'le %d %B %Y',
+            'de_DE': u'%d. %B %Y',
+            'de_CH': u'%d. %B %Y',
+            'en_US': u'%d %B %Y',
+            'it_IT': u'%d %B %Y',
+            'es_ES': u'%d de %B de %Y'
+        }
+        today = datetime.today()
+        city = _("Yverdon")
```
ecino: Should be "Yverdon-les-Bains"

ecino (changes requested):

ecino (approved):

report_compassion/models/res_partner.py
```diff
@@ -8,9 +8,11 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
-from datetime import date
+from locale import setlocale
```
ecino, line 11: This was wrong. You should import setlocale function from there
`from odoo.addons.thankyou_letters.models.res_partner import setlocale`

PR #714, opened 2018-12-07
Co2225 Account.Invoice are not updated well after a donation

nfluckiger:
This is a resolve for another bug, who would have done the same thing.


ecino:
You have some weird commits that were introduced in your branch. Please rebase onto devel and keep only the relevent commit CO-2225. Thank you.

ecino (approved):

PR #729, opened 2018-12-19
CO-2200 medical survey should be automatically completed

quentingigon:


muskathlon/models/survey_user_input.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        Automatically complete Medical Survey task when user filled it
+        """
+        res = super(SurveyUserInput, self).write(vals)
+        if vals.get('state') == 'done':
+
+            # Search for medical surveys
+            for registration in self.env['event.registration'].search([
+                ('medical_survey_id', 'in', self.ids),
+                ('partner_id_id', '=', self.partner_id.id)
```
ecino: Remove this condition, as the survey will always be linked to the same partner, plus it's illegal to do `self.partner_id` because self is a recordset (containing multiple records).

ecino: Add another condition instead checking that the event type is muskathlon.

muskathlon/models/survey_user_input.py
```diff
@@ -0,0 +1,39 @@
...
+    @api.multi
+    def write(self, vals):
+        """
+        Automatically complete Medical Survey task when user filled it
+        """
+        res = super(SurveyUserInput, self).write(vals)
+        if vals.get('state') == 'done':
+
+            # Search for medical surveys
+            for registration in self.env['event.registration'].search([
```
ecino: You don't need a for loop, as you will only call the write method. You can do instead

```
registrations = self.env[...].search(...)
registrations.write({..})
```

muskathlon/models/survey_user_input.py
```diff
@@ -0,0 +1,39 @@
...
+                ('medical_survey_id', 'in', self.ids),
+                ('partner_id_id', '=', self.partner_id.id)
+            ]):
+                # if conditions to check that the task is the correct one,
+                # same as survey in event definition
+                registration.write({
+                    'completed_task_ids': [
+                        (4, self.env.ref(
+                            'website_event_compassion.task_medical_survey').id)
```
ecino: This is the wrong task. It should be `muskathlon.task_medical` as defined in event_registration_task.xml file

ecino (changes requested):

muskathlon/models/survey_user_input.py
```diff
@@ -23,17 +23,18 @@ def write(self, vals):
...
-                # if conditions to check that the task is the correct one,
-                # same as survey in event definition
-                registration.write({
-                    'completed_task_ids': [
-                        (4, self.env.ref(
-                            'website_event_compassion.task_medical_survey').id)
-                    ]
-                })
+                ('event_id.event_type_id', '=', self.env.ref(
+                    'muskathlon.event_type_muskathlon'))
```
ecino: .id after ref to compare two ids

ecino (changes requested):

muskathlon/models/survey_user_input.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
```
ecino, line 1: This new file should be imported in __init__.py, otherwise it won't be effective.

muskathlon/models/survey_user_input.py
```diff
@@ -0,0 +1,40 @@
...
+    def write(self, vals):
+        """
+        Automatically complete Medical Survey task when user filled it
+        """
+        res = super(SurveyUserInput, self).write(vals)
+        if vals.get('state') == 'done':
+
+            # Search for medical surveys
+            registrations = self.env['event.registration'].search([
+                ('medical_survey_id', 'in', self.ids),
```
ecino: I'm sorry but after thinking of it, this won't work because the Muskathlon medical survey is not linked to the registration object. So you should link it like it was done for the regular event and survey.

In object event_registration, you should override method `next_stage` and there search for the survey and link it. We can discuss it together if you need more explanations how to do it.

ecino (changes requested):

muskathlon/models/survey_user_input.py
```diff
@@ -0,0 +1,42 @@
...
+    def write(self, vals):
+        """
+        Automatically complete Medical Survey task when user filled it
+        """
+        res = super(SurveyUserInput, self).write(vals)
+        if vals.get('state') == 'done':
+
+            # Search for medical surveys
+            registrations = self.env['event.registration'].search([
+                ('partner_id_id', 'in', self.mapped('partner_id_id')),
```
ecino: `self.mapped('partner_id_id')` doesn't exit in surveys. It should be `self.mapped('partner_id').ids`

ecino: Plus you cannot search over related fields I think. So better to put this:
```suggestion
                ('partner_id', 'in', self.mapped('partner_id').ids),
```

ecino (changes requested):

muskathlon/models/survey_user_input.py
quentingigon: 'partner_id' is not a field of event.registration, so I can't use it in the search, no ?

ecino: Yes it's a field in event.registration, only it's defined in the core module event of Odoo, not in our custom module. But you can use it, it exists.

quentingigon: Ok! thanks

quentingigon:
Sorry for bad commit message

ecino (approved):

PR #737, opened 2019-01-08
CO-2260 Mass mailing don't send to duplicate addresses

quentingigon:


mass_mailing_switzerland/models/mass_mailing.py
```diff
@@ -86,14 +86,19 @@ def send_mail(self):
             )
             emails += super(MassMailing, mailing).send_mail()
 
+        emails_set = {}
```
ecino: Here you are creating a dictionary, not a set. You should use `set()` if you want an empty set.

mass_mailing_switzerland/models/mass_mailing.py
```diff
@@ -86,14 +86,19 @@ def send_mail(self):
...
         for email in emails:
             # Only update mass mailing state when last e-mail is sent
             mass_mailing_ids = False
             if email == emails[-1]:
                 mass_mailing_ids = self.ids
 
-            # Used for Sendgrid -> Send e-mails in a job
-            email.with_delay().send_sendgrid_job(mass_mailing_ids)
+            # average complexity is O(1), so it is efficient
+            if not emails_set.__contains__(email.email_to):
```
ecino: Wouldn't it better to use the `in` operator instead of calling the private implementation of `contains` method?

ecino (changes requested):

mass_mailing_switzerland/models/mass_mailing.py
quentingigon: It would, it would !

ecino (approved):

PR #745, opened 2019-01-14
CO-2169 Add feedback survey creation

nfluckiger:


website_event_compassion/models/event_registration.py
```diff
@@ -345,6 +345,10 @@ def write(self, vals):
             for registration in self:
                 if not registration.incomplete_task_ids:
                     registration.next_stage()
+        if 'stage_id' in vals and vals['stage_id'] == self.env.ref(
+                'website_event_compassion.stage_all_attended').id:
+            if self.event_id.event_type_id.name == u'Group visit':
```
ecino: This should be inside the for loop and you should use `registration` instead of `self`, because `self` may be several records and `self.event_id` is not allowed.

ecino (changes requested):

ecino (approved):

PR #752, opened 2019-01-23
CO-2216 : Notifier Olivier des inscriptions Muskathlon

c-meier:
Allow subscription to an event_type and an event_event so that a user can be notified when a new registration is created.

The subscription to the event_type create an automated subscription for all new event_event of the given type.

### DIFF
--- a/website_event_compassion/data/event_message_subtype.xml
+++ b/website_event_compassion/data/event_message_subtype.xml
@@ -9,7 +9,7 @@
         </record>
 
         <record id="mt_event_s_registration" model="mail.message.subtype">
-            <field name="name">All new registration</field>
+            <field name="name">All new registrations</field>
             <field name="res_model">event.event</field>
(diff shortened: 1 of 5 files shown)


website_event_compassion/data/event_message_subtype.xml
```diff
@@ -0,0 +1,31 @@
...
+            <field name="res_model">event.event</field>
+            <field name="default" eval="True"/>
+            <field name="description">New registration</field>
+            <field name="sequence">10</field>
+            <field name="parent_id" eval="ref('mt_registration_create')"/>
+            <field name="relation_field">event_id</field>
+        </record>
+
+        <record id="mt_type_s_event_s_registration" model="mail.message.subtype">
+            <field name="name">Type's event's new registrations</field>
```
ecino: I suggest this name: "New registrations of event type"

ecino (approved):

PR #767, opened 2019-02-07
CO-2293 : Attacher le document child protection à la communication New Translator

c-meier:
The charter that must be agreed upon look like this:

capture d ecran de 2019-02-07 12-32-27

And this:

capture d ecran de 2019-02-07 12-32-42

The last bit must be translated.

And once agreed this page is shown (it must also be translated):

capture d ecran de 2019-02-07 12-31-42

The link (/partner/${partner.uuid}/child-protection-charter) to the charter is put in the email template, but it must be manually updated through odoo (the noupdate attribute is set).

The new fields on the res.partner form look like this:

capture d ecran de 2019-02-06 15-04-31

Be wary, I had a merge conflict when I rebased with devel: I think I correctly moved the correction to the translation of the Child Protection Charter. But there can be some problem.


partner_compassion/controllers/__init__.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2018 Compassion CH (http://www.compassion.ch)
```
ecino: You can correct your headers with year 2019 and your name as author.

ecino (changes requested):
By reading the code, I believe some behaviour is missing (but maybe I just didn't see where it's supposed to happen). Can you verify and confirm that the translator is activated on the platform when he has signed the charter?
Other than that, it seems good except for some headers that are not up to date.

sbc_switzerland/models/partner.py
```diff
@@ -39,3 +41,13 @@ def write(self, vals):
...
         return res
+
+    @api.multi
+    def agree_to_child_protection_charter(self):
+        res = super(ResPartner, self).agree_to_child_protection_charter()
+        translation = self.env.ref('partner_compassion.engagement_translation')
+        for partner in self:
+            advocate = partner.advocate_details_id
+            if translation in advocate.engagement_ids:
+                advocate.set_active()
```
c-meier: Yes, I confirm that the partner is activated when he agrees to the charter.

c-meier, line 52: Yes, I confirm that the partner is activated when he agrees to the child protection charter

ecino: :+1:

ecino (approved):

PR #771, opened 2019-02-08
CO-2269 Dissocier la communication Welcome Muskathlon d'un nouveau Portal User

jo-bloiz-ornot:
Wizard that will let you choose if you want to create a communication and let you choose which template (for now there is only Muskathlon template)

### DIFF
--- a/partner_communication_switzerland/wizards/portal_wizard.py
+++ b/partner_communication_switzerland/wizards/portal_wizard.py
@@ -48,18 +48,9 @@ class PortalWizardUser(models.TransientModel):
 
         self.mapped('partner_id').signup_prepare()
 
-        if self.env.context['create_communication']:
+        if self.env.context.get('create_communication'):
             for wizard_line in self:
                 wizard_line.create_uid_communication()
-        else:
-            for wizard_line in self:
-                p = wizard_line.partner_id
-                p.signup_prepare()
-                values = {
-                    'login': p.firstname[:1] + p.lastname,
-                    'password': 'to_change1400'
-                }
-                self.env['res.users'].signup(values, p.signup_token)
 
         return res
 
--- a/partner_communication_switzerland/wizards/portal_wizard.py
+++ b/partner_communication_switzerland/wizards/portal_wizard.py
@@ -70,4 +70,4 @@ class PortalWizardUser(models.TransientModel):
                 'partner_id': self.user_id.partner_id.id,
                 'object_ids': self.user_id.id,
                 'config_id': self.invitation_config_id.id
-        })
+            })
--- a/partner_communication_switzerland/wizards/portal_wizard.py
+++ b/partner_communication_switzerland/wizards/portal_wizard.py
@@ -9,25 +9,64 @@
 #
 ##############################################################################
 
-from odoo import api, models
+import logging
+
+from odoo import api, fields, models
+from odoo.tools.translate import _
+from odoo.exceptions import UserError
+
+
+_logger = logging.getLogger(__name__)
 
 
 class PortalWizard(models.TransientModel):
     _inherit = 'portal.wizard'
 
+    invitation_config_id = fields.Many2one('partner.communication.config')
+
+    @api.onchange('portal_id')
+    def onchange_portal_id(self):
+        # set the values of the users created in the super method
+        res = super(PortalWizard, self).onchange_portal_id()
+
+        self.user_ids.write({
+            'invitation_config_id': self.invitation_config_id.id
+        })
+
+        return res
+
+
+class PortalWizardUser(models.TransientModel):
+    _inherit = 'portal.wizard.user'
+
+    invitation_config_id = fields.Many2one('partner.communication.config')
+    uid_communication_id = fields.Many2one('partner.communication.job')
+
     @api.multi
     def action_apply(self):
-        """
-        Generate communication for User invitation
-        """
-        res = super(PortalWizard, self).action_apply()
-        comm_obj = self.env['partner.communication.job']
-        config = self.env.ref(
-            'partner_communication_switzerland.portal_welcome_config')
-        for user in self.mapped('user_ids.user_id'):
-            comm_obj.create({
-                'partner_id': user.partner_id.id,
-                'object_ids': user.id,
-                'config_id': config.id
-            })
+        res = super(PortalWizardUser, self).action_apply()
+
+        self.mapped('partner_id').signup_prepare()
+
+        # if the context is not given, create a communication
+        # else do as context say
+        if self.env.context['create_communication'] is None or \
+                self.env.context['create_communication']:
+            for wizard_line in self:
+                wizard_line.create_uid_communication()
+
         return res
+
+    @api.multi
+    def create_uid_communication(self):
+        """create a communication that contain a login url"""
+        if not self.env.user.email:
+            raise UserError(_('You must have an email address in'
+                              ' your User Preferences to send emails.'))
+        self.ensure_one()
+
+        self.uid_communication_id = self.env['partner.communication.job'].create({
+            'partner_id': self.user_id.partner_id.id,
+            'object_ids': self.user_id.id,
+            'config_id': self.invitation_config_id.id
+        })
--- a/partner_communication_switzerland/wizards/portal_wizard.py
+++ b/partner_communication_switzerland/wizards/portal_wizard.py
@@ -48,12 +48,18 @@ class PortalWizardUser(models.TransientModel):
 
         self.mapped('partner_id').signup_prepare()
 
-        # if the context is not given, create a communication
-        # else do as context say
-        if self.env.context['create_communication'] is None or \
-                self.env.context['create_communication']:
+        if self.env.context['create_communication']:
             for wizard_line in self:
                 wizard_line.create_uid_communication()
+        else:
+            for wizard_line in self:
+                p = wizard_line.partner_id
+                p.signup_prepare()
+                values = {
+                    'login': p.firstname[:1] + p.lastname,
+                    'password': 'to_change1400'
+                }
+                self.env['res.users'].signup(values, p.signup_token)
 
         return res
 
--- a/partner_communication_switzerland/wizards/portal_wizard.py
+++ b/partner_communication_switzerland/wizards/portal_wizard.py
@@ -65,8 +65,9 @@ class PortalWizardUser(models.TransientModel):
                               ' your User Preferences to send emails.'))
         self.ensure_one()
 
-        self.uid_communication_id = self.env['partner.communication.job'].create({
-            'partner_id': self.user_id.partner_id.id,
-            'object_ids': self.user_id.id,
-            'config_id': self.invitation_config_id.id
+        self.uid_communication_id = \
+            self.env['partner.communication.job'].create({
+                'partner_id': self.user_id.partner_id.id,
+                'object_ids': self.user_id.id,
+                'config_id': self.invitation_config_id.id
         })
--- a/partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
+++ b/partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2019 Compassion CH (http://www.compassion.ch)
+#    Releasing children from poverty in Jesus' name
+#    @author: Joel Vaucher <jvaucher@compassion.ch>
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+
+
+from odoo import api, models, fields, _
+
+
+class ResPartnerCreatePortalWizard(models.TransientModel):
+    """ creation of a portal user wizard will send a email
+        with the identifier if the user used the checkbox """
+    _name = 'res.partner.create.portal.wizard'
+    _description = 'take a partner and make it a odoo user'
+
+    create_communication = fields.Boolean(
+        'Send an e-mail invitation', default=True)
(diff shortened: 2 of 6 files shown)


partner_communication_switzerland/wizards/portal_wizard.py
```diff
@@ -9,25 +9,60 @@
...
-                'object_ids': user.id,
-                'config_id': config.id
-            })
+        res = super(PortalWizardUser, self).action_apply()
+
+        # to not create a communication
+        # you should give context
+        # { 'create_uid_communication': False }
+        # else it will create a communication
+        if self.env.context['create_communication_with_uid'] is None:
```
ecino: Instead of reading this value from the context, it's better and more clear to directly add a field in the wizard 'portal.wizard.user' model. `invitation_config_id = fields.Many2one('partner.communication.config')`. With that you can here directly test `if self.invitation_communication_id:`

partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
```diff
@@ -0,0 +1,80 @@
...
+          'the login link to the user ?'), default=True)
+
+    config_id = fields.Many2one(
+        'partner.communication.config', _('choose a communication config'),
+        domain="[('model', '=', 'res.users')]"
+    )
+
+    @api.multi
+    def button_create_portal_user(self):
+        portal = self.env['portal.wizard'].create({})
```
ecino: If you add the field `invitation_config_id` in the 'portal.wizard.user' model, you can here pass it in the creation values `{'invitation_config_id': self.config_id}`

partner_communication_switzerland/wizards/portal_wizard.py
```diff
@@ -9,25 +9,60 @@
...
+
         return res
+
+    @api.multi
+    def _create_uid_communication(self):
+        if not self.env.user.email:
+            raise UserError(_('You must have an email address in'
+                              ' your User Preferences to send emails.'))
+
+        config_id_id = self.env.context['config_id_id'] or \
```
ecino: If you add the field, here the communication would be simply `self.invitation_config_id`. I don't think that you should catch a case where the user wants to send something but doesn't select any communication rule. It makes no sense, either they choose one to send or we don't send nothing.

partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class ResPartnerCreatePortalWizard(models.TransientModel):
+    """ creation of a portal user wizard will send a email
+        with the identifier if the user used the checkbox """
+    _name = 'res.partner.create.portal.wizard'
+    _description = 'take a partner and make it a odoo user'
+
+    send_communication = fields.Boolean(
+        _('do you want to send a mail with '
```
ecino: Field strings don't need to be translated with the python function, they are automatically included in translations. Try to keep them short or it will become too long in the user interface. Here I propose simply `send_communication = fields.Boolean('Send an e-mail invitation')`

partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
```diff
@@ -0,0 +1,80 @@
...
+          'the login link to the user ?'), default=True)
+
+    config_id = fields.Many2one(
+        'partner.communication.config', _('choose a communication config'),
```
ecino: 'Choose a communication'

partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
```diff
@@ -0,0 +1,80 @@
...
+        for user in no_mail:
+            partner = user.partner_id
+            user.email = partner.firstname[0].lower() + \
+                partner.lastname.lower() + '@cs.local'
+
+        # define if needed to create a communication
+        bool_communication_with_uid = \
+            (self.send_communication and self.config_id)
+
+        ctx = {
```
ecino: No need to put a custom context if you pass the config via the field

partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
```diff
@@ -0,0 +1,80 @@
...
+
+    config_id = fields.Many2one(
+        'partner.communication.config', _('choose a communication config'),
+        domain="[('model', '=', 'res.users')]"
+    )
+
+    @api.multi
+    def button_create_portal_user(self):
+        portal = self.env['portal.wizard'].create({})
+        portal.onchange_portal_id()
```
ecino, line 37: Add a comment here to explain that the following lines are for allowing registering users even if they don't have an e-mail address.

partner_communication_switzerland/wizards/portal_wizard.py
```diff
@@ -9,25 +9,60 @@
...
+            user_id = wizard_line.user_id
+            partner_id = user_id.partner_id
+
+            self.env['partner.communication.job'].create({
+                'partner_id': partner_id.id,
+                'object_ids': user_id.id,
+                'config_id': config_id_id
+            })
+
+        return True
```
ecino: Instead of returning True, you can return the communication created in order to use it directly. `return self.env['partner.communication.job'].create`

partner_communication_switzerland/wizards/portal_wizard.py
```diff
@@ -9,25 +9,60 @@
...
+        # { 'create_uid_communication': False }
+        # else it will create a communication
+        if self.env.context['create_communication_with_uid'] is None:
+            bool_communication_with_uid = True
+        else:
+            bool_communication_with_uid = \
+                self.env.context['create_communication_with_uid']
+
+        if bool_communication_with_uid:
+            self._create_uid_communication()
```
ecino: Here you can add inside res the communication created (or just the id maybe if it doesn't work). `res['communication'] = self._create_uid_communication()`. `_create_uid_communication` has of course to return the communication record.

partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
```diff
@@ -0,0 +1,80 @@
...
+            'create_communication_with_uid': bool_communication_with_uid,
+            'config_id_id': self.config_id.id
+        }
+        portal.with_context(ctx).action_apply()
+        no_mail.mapped('partner_id').write({'email': False})
+
+        action = None
+        if bool_communication_with_uid:
+            # get same communication create in portal.action_apply()
+            same_job_search = [
```
ecino: If your function `action_apply` puts the communication inside the result, here you can simply retrieve it by doing `communication = action_result['communication']`

partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # define if needed to create a communication
+        bool_communication_with_uid = \
+            (self.send_communication and self.config_id)
+
+        ctx = {
+            'create_communication_with_uid': bool_communication_with_uid,
+            'config_id_id': self.config_id.id
+        }
+        portal.with_context(ctx).action_apply()
```
ecino: assign the result in order to retrieve the generated communication : `action_result = portal.action_apply()`

ecino (changes requested):
Very good changes, thank you! Here are some suggestions in order to improve how parameters and results are passed between the different objects and methods.

partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
jo-bloiz-ornot: The problem is that it's not a portal wizard user but just a portal wizard, I think overriding the method (onchange_id) that create portal user is possible but it's will lose the adaptability of the oca module

jo-bloiz-ornot: I think I can do it without losing changes of oca module, I tell you later

partner_communication_switzerland/wizards/portal_wizard.py
```diff
@@ -9,25 +9,73 @@
...
-        config = self.env.ref(
-            'partner_communication_switzerland.portal_welcome_config')
-        for user in self.mapped('user_ids.user_id'):
-            comm_obj.create({
-                'partner_id': user.partner_id.id,
-                'object_ids': user.id,
-                'config_id': config.id
-            })
-        return res
+        self.ensure_one()
```
ecino: You removed the call to the super which here can remove some useful features. Can you add it back?

ecino (changes requested):

partner_communication_switzerland/wizards/portal_wizard.py
jo-bloiz-ornot: No, because the parent didn't return the result of the action_apply method

```
    @api.multi
    def action_apply(self):
        self.ensure_one()
        self.user_ids.action_apply()
        return {'type': 'ir.actions.act_window_close'}
```
Override had to be done there

ecino: The problem is that this code will never be executed :
https://github.com/compassionch/compassion-modules/blob/bac7e809eb6deabea16c080698f84122601a69d9/crm_compassion/wizards/portal_wizard.py#L21
You should call the super even if it doesn't return what we need and then return what you need after. Maybe here some conception needs to be changed to ensure we don't lose the code of crm_compassion.

jo-bloiz-ornot: ok

Just one question, is the fact that the method portal.wizard.user action_apply() will be call once in the super method then again just to get the result is a problem ?

jo-bloiz-ornot: I will try with a additional field in portal.wizard.user to not call twice the method

jo-bloiz-ornot: it's a success

partner_communication_switzerland/wizards/portal_wizard.py
```diff
@@ -9,25 +9,65 @@
...
-                'object_ids': user.id,
-                'config_id': config.id
-            })
+        res = super(PortalWizardUser, self).action_apply()
+
+        self.mapped('partner_id').signup_prepare()
+
+        # if the context is not given, create a communication
+        # else do as context say
+        if self.env.context['create_communication'] is None or \
```
ecino: I don't understand why you create a communication when context is None. To me, it sounds more logic to create the communication only when we are told to do so. This would mean `if self.env.context.get('create_communication'):`

partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
```diff
@@ -0,0 +1,71 @@
...
+            user.email = partner.firstname[0].lower() + \
+                partner.lastname.lower() + '@cs.local'
+
+        portal.with_context({
+            'create_communication': self.create_communication
+        }).action_apply()
+
+        no_mail.mapped('partner_id').write({'email': False})
+
+        action = None
```
ecino: `action = True` is better to return.

partner_communication_switzerland/wizards/res_partner_create_portal_wizard.py
```diff
@@ -0,0 +1,71 @@
...
+                'view_type': 'form',
+                'view_mode': 'tree,form',
+                'domain': [('id', 'in', uid_communication.ids)],
+                'flags': {'action_buttons': True}
+            }
+        return action
+
+    @api.multi
+    def button_cancel(self):
+        return None
```
ecino: return True (means ok).

partner_communication_switzerland/wizards/portal_wizard.py
```diff
@@ -9,25 +9,71 @@
...
-            comm_obj.create({
-                'partner_id': user.partner_id.id,
-                'object_ids': user.id,
-                'config_id': config.id
-            })
+        res = super(PortalWizardUser, self).action_apply()
+
+        self.mapped('partner_id').signup_prepare()
+
+        if self.env.context['create_communication']:
```
ecino:
```suggestion
        if self.env.context.get('create_communication'):
```
This avoids the code execution to fail in case `create_communication` is not present in the context

partner_communication_switzerland/wizards/portal_wizard.py
```diff
@@ -9,25 +9,71 @@
...
+            for wizard_line in self:
+                wizard_line.create_uid_communication()
+        else:
+            for wizard_line in self:
```
ecino: Why did you add this? I'm not sure we want to automatically activate the account if we don't send a communication.

ecino (changes requested):

partner_communication_switzerland/wizards/portal_wizard.py
jo-bloiz-ornot: good to know, thank you

partner_communication_switzerland/wizards/portal_wizard.py
jo-bloiz-ornot: I was afraid to give to much work to the user to activate manually the account, because if you dont send a email with a link to activate the account, who gonna activate it ?

ecino: In fact we register some users that we never want to give access to the account, only to be able to assign them to some events or other objects (but this is intern stuff and we don't communicate to them). So we don't wan't that everyone is activated automatically. I think you can remove this part.

jo-bloiz-ornot: alright, it's gonna be done

ecino (approved):

PR #785, opened 2019-02-19
CO-2337 deleted all refs to partner titles different than Mister and Madam (or Family)

quentingigon:


ecino (changes requested):
Great, but I need a migration script that transfers the partner that had these titles to the correct equivalent (Mister or Madam).

partner_compassion/migrations/10.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,26 @@
...
+        return
+
+    # id value for Doctors, Pastors, Professor, Miss
+    # records don't exist anymore so we don't use env.ref to get their ids
+    bad_ids = [7, 30, 45, 58]
+
+    for partner in env['res.partner'].search([
+        ("title.id", "in", bad_ids)
+    ]):
+        partner.title = env.ref('base.res_partner_title_mister').id
```
ecino: Miss must become "Madam" and not "Mister"

ecino (changes requested):

ecino (approved):

PR #801, opened 2019-02-26
CO-2391 : reuse partner matching object for wordpress sponsorship

c-meier:
Need to be done after CompassionCH/compassion-modules#943


ecino (changes requested):
Code looks good to me but there were some automated tests on the module that are broken and need to be adapted. This is the good time for me to show you how the unit tests are done in Odoo. When you have time we can look at this. Here are the tests results on travis: https://travis-ci.org/CompassionCH/compassion-switzerland/jobs/498784281

PR #802, opened 2019-02-27
CO-2241 Ajouter un mécanisme pour suivre les parrainages en attente de mandat

jo-bloiz-ornot:
add a field mandate_date for checking if you wait for more than 3 month in mandate state

### DIFF
--- a/sponsorship_switzerland/views/res_partner_view.xml
+++ b/sponsorship_switzerland/views/res_partner_view.xml
@@ -10,4 +10,15 @@
             </field>
         </field>
     </record>
+
+    <record id="action_view_partner_waiting_mandate" model="ir.actions.server" >
+        <field name="name">Waiting mandate</field>
+        <field name="model_id" ref="model_recurring_contract"/>
+        <field name="code">
+action = model.action_view_waiting_mandate()
+        </field>
+    </record>
+
+    <menuitem id="menu_waiting_mandate" parent="child_compassion.menu_sponsorship_section"
+              name="Waiting mandate" action="action_view_partner_waiting_mandate" sequence="100"/>
 </odoo>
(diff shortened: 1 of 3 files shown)


jo-bloiz-ornot:
maybe 3 month is too long ?

ecino:
> maybe 3 month is too long ?

Yes it's too long. 3 weeks is better.

sponsorship_switzerland/views/res_partner_view.xml
```diff
@@ -10,4 +10,15 @@
             </field>
         </field>
     </record>
+
+    <record id="action_view_partner_waiting_mandate" model="ir.actions.server" >
```
ecino: It's better to define a window action, with that you will not even need any method in recurring.contract object!
```suggestion
    <record id="action_view_partner_waiting_mandate" model="ir.actions.act_window">
        <field name="name">Waiting mandate</field>
        <field name="res_model">recurring.contract</field>
        <field name="view_type">form</field>
        <field name="view_mode">tree,form</field>
        <field name="domain" eval="[('state', '=', 'mandate'), ('mandate_date', '<', (DateTime.now() - timedelta(days=25)).strftime('%Y-%m-%d'))]"/>
    </record>
```

ecino (changes requested):
One flake8 issue in migration:
`sponsorship_switzerland/migrations/10.0.1.4.0/post-migration.py:20:31: W291 trailing whitespace`

sponsorship_switzerland/views/res_partner_view.xml
jo-bloiz-ornot: that would take me a loong time to find, thank you

ecino (approved):

PR #803, opened 2019-02-27
CO-2388 Force cancel invoices at the end of a sponsorship

quentingigon:
It doesn't work on my local database, because of the method "cancel_payment_lines()", where it can't find a payment_line that has the same move_line_id as the move associated with the invoice


sponsorship_switzerland/models/contracts.py
```diff
@@ -411,3 +412,14 @@ def _reconcile_open_amount(self):
             delay = datetime.now() + relativedelta(seconds=15)
             if invoices:
                 invoices.with_delay(eta=delay).group_or_split_reconcile()
+
+    @api.multi
+    @job(default_channel='root.recurring_invoicer')
+    @related_action(action='related_action_contract')
+    def _clean_invoices(self, since_date=None, to_date=None, keep_lines=None):
+        inv_lines = self.invoice_line_ids.filtered(lambda r: r.state == 'open')
+        for invoice in inv_lines.mapped('invoice_id'):
```
ecino: There is no need to loop. You can simply call `inv_lines.cancel_payment_lines()`

sponsorship_switzerland/models/contracts.py
```diff
@@ -411,3 +412,14 @@ def _reconcile_open_amount(self):
             delay = datetime.now() + relativedelta(seconds=15)
             if invoices:
                 invoices.with_delay(eta=delay).group_or_split_reconcile()
+
+    @api.multi
+    @job(default_channel='root.recurring_invoicer')
+    @related_action(action='related_action_contract')
+    def _clean_invoices(self, since_date=None, to_date=None, keep_lines=None):
+        inv_lines = self.invoice_line_ids.filtered(lambda r: r.state == 'open')
```
ecino: Actually you need to include paid invoices that are in the future, in case the sponsor paid in advance, because those invoices will as well be cancelled.

ecino (changes requested):

ecino:
`     self._clean_invoices(since_date, to_date, keep_lines)
`   File "/home/travis/build/CompassionCH/compassion-switzerland/sponsorship_switzerland/models/contracts.py", line 422, in _clean_invoices
`     today = datetime.datetime.now()
` AttributeError: type object 'datetime.datetime' has no attribute 'datetime'

ecino:
Now there are some conflicts with other changes that I merged. Can you please look at it?

quentingigon:
I have not yet looked for datetime

sponsorship_switzerland/models/contracts.py
```diff
@@ -413,3 +414,32 @@ def _reconcile_open_amount(self):
...
+    @api.multi
+    @job(default_channel='root.recurring_invoicer')
+    @related_action(action='related_action_contract')
+    def _clean_invoices(self, since_date=None, to_date=None, keep_lines=None):
+        today = datetime.now()
+        inv_lines = self.invoice_line_ids.filtered(lambda r: r.state == 'open'
+                                                   or (r.state == 'paid'
+                                                      and r.date > today))
+
+        for line in inv_lines:
```
ecino: You went back to a previous code version. Here you can do `inv_lines.cancel_payment_lines()`

sponsorship_switzerland/models/contracts.py
```diff
@@ -413,3 +414,32 @@ def _reconcile_open_amount(self):
...
                 invoices.with_delay(eta=delay).group_or_split_reconcile()
+
+    @api.multi
+    @job(default_channel='root.recurring_invoicer')
+    @related_action(action='related_action_contract')
+    def _clean_invoices(self, since_date=None, to_date=None, keep_lines=None):
+        today = datetime.now()
+        inv_lines = self.invoice_line_ids.filtered(lambda r: r.state == 'open'
+                                                   or (r.state == 'paid'
+                                                      and r.date > today))
```
ecino: r.date is a string and today is a datetime. You should do `fields.Datetime.from_string(r.date) > today`

sponsorship_switzerland/models/contracts.py
```diff
@@ -413,3 +414,32 @@ def _reconcile_open_amount(self):
             delay = datetime.now() + relativedelta(seconds=15)
             if invoices:
                 invoices.with_delay(eta=delay).group_or_split_reconcile()
+
+    @api.multi
+    @job(default_channel='root.recurring_invoicer')
+    @related_action(action='related_action_contract')
+    def _clean_invoices(self, since_date=None, to_date=None, keep_lines=None):
+        today = datetime.now()
```
ecino: now() is a datetime, better to use a date: `datetime.today()`

sponsorship_switzerland/models/contracts.py
```diff
@@ -413,3 +414,32 @@ def _reconcile_open_amount(self):
...
+                                                      and r.date > today))
+
+        for line in inv_lines:
+            line.invoice_id.cancel_payment_lines()
+
+        return super(RecurringContracts, self)._clean_invoices(
+            since_date, to_date, keep_lines)
+
+    @api.model
+    def _needaction_count(self, domain=None):
```
ecino: This function has nothing to do with your pull request. Please remove this code.

ecino (changes requested):

sponsorship_switzerland/models/contracts.py
quentingigon: Yes, I made a mistake during the merge with devel

quentingigon:
The checks are failing but I don't understand why, as the message is that 'account.invoice' doesn't have a function called cancel_payment_lines(). But it does..

ecino:
That's because the module sponsorship_switzerland doesn't depend on account_payment_line_cancel module. Please add that dependency in the manifest.

PR #810, opened 2019-03-06
CO-2401 created new informal salutation for our donators

quentingigon:


partner_communication_switzerland/models/res_partner.py
```diff
@@ -45,6 +46,7 @@ def _compute_salutation(self):
         for p in company:
             p.salutation = _("Dear friends of compassion")
             p.short_salutation = p.salutation
+            p.informal_salutation = _("Dear" + self._name)
```
ecino, line 49: self._name will return 'res.partner'. Please test your changes before doing pull requests :)

You should use the title to find the correct "Dear", could be "Cher" or "Chère" or "Chers" and p.firstname to get the firstname of the partner.

ecino (changes requested):

PR #812, opened 2019-03-06
CO-2241 add number beside menuitem for people waiting mandate for more than 20 days

jo-bloiz-ornot:


jo-bloiz-ornot:
almost a clean code, the bad part is that all menu that have the same related object AND same domain as the action waiting_mandate will also have the number

ecino (approved):

PR #818, opened 2019-03-08
CO-2404 : Child protection charter for muskathlon

c-meier:
There is some translation that needs to be done.

### DIFF
--- a/muskathlon/__manifest__.py
+++ b/muskathlon/__manifest__.py
@@ -62,7 +62,7 @@
         'templates/muskathlon_order_material.xml',
         'templates/assets.xml'
     ],
-    'depends': ['website_event_compassion', 'survey'],
+    'depends': ['website_event_compassion', 'survey', 'partner_compassion'],
     'external_dependencies': {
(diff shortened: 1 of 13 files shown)


muskathlon/__manifest__.py
```diff
@@ -62,7 +62,7 @@
         'templates/muskathlon_order_material.xml',
         'templates/assets.xml'
     ],
-    'depends': ['website_event_compassion', 'survey'],
+    'depends': ['website_event_compassion', 'survey', 'partner_compassion'],
```
ecino, line 65: website_event_compassion is already depending on partner_compassion (through the following chain dependency: website_event_compassion -> partner_communication_switzerland -> report_compassion -> sponsorship_switzerland -> partner_compassion.

Thus, you don't need to add this direct dependency.

ecino (changes requested):
Just one dependency not needed, otherwise the code is good.

ecino:
Add child protection charter for Muskathlon

PR #825, opened 2019-03-15
Co 2410 Add child information to the picture

nfluckiger:


ecino (changes requested):
Two flake8 issues:
```
partner_communication_switzerland/wizards/child_order_picture.py:23:1: E402 module level import not at top of file
partner_communication_switzerland/wizards/child_order_picture.py:104:80: E501 line too long (83 > 79 characters)
```
The travis build fails because of pdf2image library. You should list it in `requirements.txt` file.
Finally, you can increase version number of module in manifest file.

PR #835, opened 2019-03-21
Automated email response when in holiday - mail template 

quentingigon:


ecino:
The changes should be made in `crm_request` with your other changed. `crm_request` depends on partner_communication, which is fine for adding these templates there.

PR #845, opened 2019-03-28
CO-2459 Force payment mode to 'BVR' for travel cost

jonathantarabbia:


website_event_compassion/models/event_registration.py
```diff
@@ -604,6 +611,7 @@ def prepare_group_visit_payment(self):
                 'name': product.name,
                 'product_id': product.id,
                 'account_analytic_id': event.analytic_id.id,
+                'payment_mode_id': mode_pay_bvr.id,
```
ecino: This is invoice line vals and invoice line doesn't have a `payment_mode_id` field. You can thus remove this line. It should only be included in invoice values.

website_event_compassion/models/event_registration.py
```diff
@@ -617,6 +625,7 @@ def prepare_group_visit_payment(self):
...
```
ecino: This is invoice line and doesn't need payment_mode.

website_event_compassion/models/event_registration.py
```diff
@@ -629,6 +638,7 @@ def prepare_group_visit_payment(self):
             'name': product.name,
             'product_id': product.id,
             'account_analytic_id': event.analytic_id.id,
+            'payment_mode_id': mode_pay_bvr.id,
```
ecino: This is invoice line, it doesn't need payment_mode

ecino: You should add the payment_mode just a bit below, at line 639 (when creating the invoice)

ecino (changes requested):

ecino (approved):

PR #851, opened 2019-04-02
CO-2499 fix-duplicate-partner

jonathantarabbia:
- no copy field advocate_details_id
 - no copy field geo_point (crash on duplicate partner)


ecino (approved):
Great!

PR #855, opened 2019-04-04
CO-2491 Don't create alias if there no email

eicher31:


ecino:
Why is this draft, are you still working on it?

ecino (approved):

PR #856, opened 2019-04-09
CO-2517 fix tshirt size, birthdate, won't fix image orientation

vwvw:


muskathlon/migrations/10.0.3.2.0/pre-migration.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    if not version:
+        return
+
+    singlets = env['event.registration'].search(
+        [('t_shirt_type', '=', 'singlet')])
+    for reg in singlets:
```
ecino: You can replace the loop with the write method:
```suggestion
    singlets.write({'t_shirt_type': 'bikeshirt'})
```

ecino (changes requested):
Missing:
- migrate XS  t-shirts to S
- increase the version in the manifest file so that the migration can happen

ecino (approved):

PR #860, opened 2019-04-15
CO-2488 Ending sponsorship communication

jo-bloiz-ornot:
add a variation about ending
 -> if an activated sponsorship end and is not cancelled
    it's a terminated communication that is send

### DIFF
--- a/partner_communication_switzerland/models/contracts.py
+++ b/partner_communication_switzerland/models/contracts.py
@@ -529,6 +529,11 @@ class RecurringContract(models.Model):
         ).send_communication(cancellation, both=True)
         self.filtered(
             lambda s: s.end_reason != '1' and s.parent_id
+            and bool(s.activation_date)
+        ).send_communication(cancellation, correspondent=False)
+        self.filtered(
+            lambda s: s.end_reason != '1' and s.parent_id
+            and not bool(s.activation_date)
         ).send_communication(no_sub, correspondent=False)
 
     def _new_dossier(self):


jo-bloiz-ornot:
Just not sure if  "partner_communication_switzerland.planned_completion" is the good template

also I didnt know of to run a test of the reprise sponsorships to check the three options there.
```
    cancellation = self.env.ref(
    'partner_communication_switzerland.sponsorship_cancellation')
    completion = self.env.ref(
    'partner_communication_switzerland.planned_completion')
    no_sub = self.env.ref(
    'partner_communication_switzerland.planned_no_sub')
```

but the change is small, so check should be easy

partner_communication_switzerland/models/contracts.py
```diff
@@ -522,13 +522,20 @@ def _on_sponsorship_finished(self):
         super(RecurringContract, self)._on_sponsorship_finished()
         cancellation = self.env.ref(
             'partner_communication_switzerland.sponsorship_cancellation')
+        completion = self.env.ref(
+            'partner_communication_switzerland.planned_completion')
```
```suggestion
            'partner_communication_switzerland.sponsorship_cancellation')
```

partner_communication_switzerland/models/contracts.py
```diff
@@ -522,13 +522,20 @@ def _on_sponsorship_finished(self):
         super(RecurringContract, self)._on_sponsorship_finished()
         cancellation = self.env.ref(
             'partner_communication_switzerland.sponsorship_cancellation')
+        completion = self.env.ref(
```
```suggestion
        cancellation = self.env.ref(
```

partner_communication_switzerland/models/contracts.py
```diff
@@ -522,13 +522,20 @@ def _on_sponsorship_finished(self):
...
+            'partner_communication_switzerland.planned_completion')
         no_sub = self.env.ref(
             'partner_communication_switzerland.planned_no_sub')
         self.filtered(
             lambda s: s.end_reason != '1' and not s.parent_id
         ).send_communication(cancellation, both=True)
         self.filtered(
             lambda s: s.end_reason != '1' and s.parent_id
+            and bool(s.activation_date)
+        ).send_communication(completion, correspondent=False)
```
```suggestion
        ).send_communication(cancellation, correspondent=False)
```

ecino (changes requested):
Everything is good, but it's just another communication that is used.

jo-bloiz-ornot:
The cancellation communication was already fetch just above so I just change the communication used at the end

ecino (approved):

PR #862, opened 2019-04-17
CO-2493 Delete medical records once event is more than 31 days old. 

vwvw:
Additionally Fix wrong link to medical id.

### DIFF
--- a/muskathlon/migrations/10.0.3.3.0/pre-migration.py
+++ b/muskathlon/migrations/10.0.3.3.0/pre-migration.py
@@ -18,7 +18,9 @@ def migrate(env, version):
     if not version:
         return
 
-    for registration in env['event.registration'].search([]):
+    for registration in env['event.registration'].search([
+            ('survey_id', '=', False),
+            ('event_id.medical_survey_id', '!=', False)]):
         survey = registration.event_id.medical_survey_id
         medical_survey_id = env['survey.user_input'].search([
             ('partner_id', '=', registration.partner_id_id),
--- a/muskathlon/migrations/10.0.3.3.0/pre-migration.py
+++ b/muskathlon/migrations/10.0.3.3.0/pre-migration.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2019 Compassion CH (http://www.compassion.ch)
+#    Releasing children from poverty in Jesus' name
+#    @author: Nicolas Badoux <n.badoux@hotmail.com>
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    if not version:
+        return
+
+    for registration in env['event.registration'].search([]):
+        survey = registration.event_id.medical_survey_id
+        medical_survey_id = env['survey.user_input'].search([
+            ('partner_id', '=', registration.partner_id_id),
+            ('survey_id', '=', survey.id)
+        ], limit=1)
+        registration.medical_survey_id = medical_survey_id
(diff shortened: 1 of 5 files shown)


muskathlon/migrations/10.0.3.3.0/pre-migration.py
```diff
@@ -0,0 +1,27 @@
...
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    if not version:
+        return
+
+    for registration in env['event.registration'].search([]):
```
ecino: I would limit the search to registrations that have no medical survey attached.
```suggestion
    for registration in env['event.registration'].search([
            ('survey_id', '=', False),
            ('event_id.medical_survey_id', '!=', False)]):
```

muskathlon/models/event_compassion.py
```diff
@@ -85,3 +86,14 @@ def get_muskathlon_participants(self, event_id):
                 'country': participant.partner_id.country_id.name
             })
         return ret
+
+    @api.multi
```
ecino: CRON methods are static
```suggestion
    @api.model
```

muskathlon/models/event_compassion.py
```diff
@@ -85,3 +86,14 @@ def get_muskathlon_participants(self, event_id):
...
+    def _cron_delete_medical_surveys(self):
+        for event in self.env['crm.event.compassion'].search([
```
ecino:
```suggestion
        for event in self.search([
```

muskathlon/models/event_compassion.py
```diff
@@ -85,3 +86,14 @@ def get_muskathlon_participants(self, event_id):
...
+            ('end_date', '<',
+             fields.Datetime.to_string(datetime.now() - timedelta(days=31)))]):
+            for registration in event.registration_ids:
+                if len(registration.medical_survey_id) > 0:
```
ecino: No need to add this test. unlink() method will simply do nothing if the relation is empty (will return an empty recordset and unlink will do nothing).

muskathlon/models/survey_user_input.py
```diff
@@ -31,12 +31,19 @@ def write(self, vals):
                     'muskathlon.stage_fill_profile').id)
             ])
 
+            survey = registrations.event_id.medical_survey_id
```
ecino: Be careful because `registrations` could potentially be a recordset (not a single record)
```suggestion
            surveys = registrations.mapped('event_id.medical_survey_id')
```

muskathlon/models/survey_user_input.py
```diff
@@ -31,12 +31,19 @@ def write(self, vals):
...
+            medical_survey_id = self.env['survey.user_input'].search([
+                ('partner_id', '=', registrations.partner_id_id),
```
ecino: Here you will need to loop over the registrations in order to assign the correct medical survey for each registration.

ecino (changes requested):

muskathlon/models/survey_user_input.py
```diff
@@ -31,12 +31,20 @@ def write(self, vals):
...
-            # if conditions to check that the task is the correct one,
-            # same as survey in event definition
-            registrations.write({
-                'completed_task_ids': [
-                    (4, self.env.ref(
-                        'muskathlon.task_medical').id)
-                ]
-            })
+            surveys = registrations.mapped('event_id.medical_survey_id')
+            for survey in surveys:
```
ecino: I would rather loop over the registrations and take one survey at a time. Because here you will write the last found survey in all registrations.
```
            for registration in registrations:
                survey = registration.event_id.medical_survey_id
                ...
                registration.write({'completed_task_ids': ...})
```

ecino (changes requested):

ecino (approved):

PR #866, opened 2019-04-24
CO-2562 Correct a bug on the word_press connector

nfluckiger:


wordpress_connector/models/contracts.py
```diff
@@ -87,7 +87,7 @@ def create_sponsorship(self, child_local_id, form_data, sponsor_lang,
         _logger.info("New sponsorship for child %s from Wordpress: %s",
                      child_local_id, str(form_data))
         try:
-
+            form_data['Sponsorship Child'] = child_local_id
```
ecino:
```suggestion
            form_data['Child reference'] = child_local_id
```

ecino (changes requested):
Instead of forcing to English when they have no language, I would rather modify at line 103 and check :+1: 
```
            if form_data.get('language'):
                partner_infos['spoken_lang_ids'] = match_obj.match_spoken_langs(
                    form_data['language']
                )
```

ecino (approved):

wordpress_connector/models/contracts.py
```diff
@@ -100,9 +100,10 @@ def create_sponsorship(self, child_local_id, form_data, sponsor_lang,
             partner_infos['title'] = match_obj.match_title(
                 form_data['salutation']
             )
-            partner_infos['spoken_lang_ids'] = match_obj.match_spoken_langs(
-                form_data['language']
-            )
+            if form_data.get('language'):
+                partner_infos['spoken_lang_ids'] = match_obj.match_spoken_langs(
```
ecino: Line is too long :cry:

PR #868, opened 2019-04-25
CO-2564 tracing email

jo-bloiz-ornot:
update email field of res partner to be traced


ecino:
@joblo-compassion please include the listed fileds in my comment on the jira task.

PR #877, opened 2019-05-10
Co 2571 The payment methode in reconcile invoice is get in the journal 

nfluckiger:


account_reconcile_compassion/wizards/reconcile_fund_wizard.py
```diff
@@ -65,6 +65,7 @@ def reconcile_with_fund(self):
                 account_id = invoice.account_id.id
                 partner_id = line.partner_id.id
                 active_ids.remove(line.id)
+                journal = line.journal_id
```
ecino: The variable journal is never used. Are you sure the code is doing anything?

account_reconcile_compassion/views/account_invoice_view.xml
```diff
@@ -10,6 +10,7 @@
                 <button name="change_attribution" type="object" string="Reconcile" class="oe_highlight"
                         attrs="{'invisible': [('to_reconcile', '=', 0.0)]}"/>
             </xpath>
+            <xpath expr="//field[@name='team_id']" position="replace"/>
```
ecino: It's better to make the field invisible instead of removing it, to guarantee compatibility with other modules that might use the field.
```suggestion
            <xpath expr="//field[@name='team_id']" position="attributes">
                <attribute name="invisible">1</attribute>
            </xpath>
```

ecino (changes requested):

ecino (approved):

PR #880, opened 2019-05-16
CO-2527 Automatically send muskathlon's medical surveys to designated doctor

quentingigon:


quentingigon:
I forgot to include a file..

PR #881, opened 2019-05-16
CO-2570 FIX bvr reference for compassion event

eicher31:


ecino (approved):
Fix flake8 imports and it's good for me :+1:

PR #894, opened 2019-06-13
CO-2607 Set paiement term as imediate for web donation

eicher31:


eicher31:
@ecino
Is the error due to my changes?

website_event_compassion/forms/donation_form.py
```diff
@@ -128,6 +128,9 @@ def form_before_create_or_update(self, values, extra_values):
             })
 
         def _form_create(self, values):
+            # Set payment term as immediate
+            values['payment_term_id'] = self.env.ref(
```
ecino: Why not including at line 127 where we already define all invoice values?

ecino (changes requested):

ecino:
> @ecino
> Is the error due to my changes?

No it was another error that should be fixed now.

website_event_compassion/forms/donation_form.py
ecino: Could you check it?

eicher31: Yes it's done

ecino (approved):

PR #895, opened 2019-06-14
CO-2469 SMS are objects and displayed in Interaction Resume

Theyiot:
SMS, which were previously simply displayed at the bottom of the profile page of a sponsor, are now stored in a new table in the database. In addition to that, they are displayed in the "interaction resume" view, along with phone calls, e-mails, and letters.


sms_939/models/sms_sponsorship.py
```diff
@@ -0,0 +1,19 @@
...
+#    @author: Théo Nikles <theo.nikles@gmail.com>
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+from odoo import models, fields
+
+
+class SponsorshipSMS(models.Model):
+    _name = 'sms_sponsorship'
```
ecino: sms sponsorship is not a good name, because it has nothing to do with sponsorships. It's only sms logs so I would propose this name instead: `sms.log`. Please rename all needed objects.

sms_939/models/sms_sponsorship.py
```diff
@@ -0,0 +1,19 @@
...
+
+    partner_id = fields.Id(string='partner_id')
```
ecino: This should be a Many2one field on table `res.partner` string should be 'Partner' (it's used for the user interface display and should be a friendly name).

sms_939/models/sms_sponsorship.py
```diff
@@ -0,0 +1,19 @@
...
+    text = fields.Char(string='text')
```
ecino: `fields.Text` would allow more characters. String attribute is not needed when the string is the same as the field name. I would also add the `required` attribute, because it doesn't make sense to make a record without text.

sms_939/models/sms_sponsorship.py
```diff
@@ -0,0 +1,19 @@
...
+    subject = fields.Text(string='subject')
+    date = fields.Datetime(string='date')
```
ecino: string attribute not required if it's the same as the field name.

sms_939/security/ir.model.access.csv
```diff
@@ -1,5 +1,6 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 sms_notification_full,SMS Notification full access,model_sms_notification,base.group_system,1,1,1,1
 sms_hook_full,SMS hook full access,model_sms_hook,base.group_system,1,1,1,1
+sms_sponsorship_full,SMS sponsorship full access,model_sms_sponsorship,base.group_system,1,1,1,1
```
ecino: Please also add all accesses to `base.group_user` except for unlink. Employees should be able to log SMS.

sms_939/wizards/sms_sender_wizard.py
```diff
@@ -84,20 +85,24 @@ def send_sms_partner(self):
...
         if self.send_sms(self.partner_id.mobile.replace(u'\xa0', u'')):
-            self.partner_id.message_post(body=self.text,
-                                         subject=self.subject,
-                                         partner_ids=self.partner_id,
-                                         type='comment')
+            self.env['sms_sponsorship'].create({
+                'partner_id': self.partner_id.id,
+                'text': self.text,
+                'subject': self.subject,
+                'date': datetime.today()
```
ecino: `fields.Datetime.now()` will also remove the added import of the datetime library.

sms_939/wizards/sms_sender_wizard.py
```diff
@@ -84,20 +85,24 @@ def send_sms_partner(self):
...
             return False
         if self.send_sms(self.sms_request_id.sender.replace(u'\xa0', u'')):
-            self.sms_request_id.message_post(body=self.text,
-                                             subject=self.subject,
-                                             type='comment')
+            self.env['sms_sponsorship'].create({
+                'partner_id': self.partner_id.id,
+                'text': self.text,
+                'subject': self.subject,
+                'date': datetime.today()
```
ecino: `fields.Datetime.now()`

ecino (changes requested):

sms_939/wizards/sms_sender_wizard.py
Theyiot: Done.

sms_939/wizards/sms_sender_wizard.py
Theyiot: Done.

sms_939/security/ir.model.access.csv
Theyiot: Done.

sms_939/models/sms_sponsorship.py
Theyiot: Done.

sms_939/models/sms_sponsorship.py
Theyiot: Done.

sms_939/models/sms_sponsorship.py
Theyiot: Done.

sms_939/models/sms_sponsorship.py
Theyiot: Done.

sms_939/security/ir.model.access.csv
```diff
@@ -1,5 +1,7 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 sms_notification_full,SMS Notification full access,model_sms_notification,base.group_system,1,1,1,1
 sms_hook_full,SMS hook full access,model_sms_hook,base.group_system,1,1,1,1
+sms_sponsorship_full,SMS sponsorship full access,model_sms_log,base.group_system,1,1,1,1
+sms_sponsorship_user,SMS sponsorship user access,model_sms_log,base.group_user,1,1,1,1
```
ecino: Unlink access should be 0
```suggestion
sms_sponsorship_user,SMS sponsorship user access,model_sms_log,base.group_user,1,1,1,0
```

sms_939/models/sms_log.py
```diff
@@ -0,0 +1,19 @@
...
+##############################################################################
+from odoo import models, fields
+
+
+class LogSMS(models.Model):
+    _name = 'sms.log'
+
+    partner_id = fields.Many2one('res.partner', 'Partner')
+    text = fields.Text(required=True)
+    subject = fields.Text()
```
ecino:
```suggestion
    subject = fields.Char()
```

ecino (changes requested):

sms_939/security/ir.model.access.csv
Theyiot: Done.

sms_939/models/sms_log.py
Theyiot: Done.

ecino (approved):

PR #906, opened 2019-06-28
CO-2100 Update Sms_sponsorship to use website child if it has no event

nfluckiger:


sms_939/models/sms_child_request.py
```diff
@@ -41,3 +43,47 @@ def send_step1_reminder(self):
             })
             sms_sender.send_sms_request()
         super(SmsRequest, self).send_step1_reminder()
+
+    @job(default_channel='root.sms_request')
+    @related_action(action='related_action_sms_request')
+    def reserve_child(self):
+
```
ecino:
```suggestion
        self.ensure_one()
```

sms_939/models/sms_child_request.py
```diff
@@ -41,3 +43,47 @@ def send_step1_reminder(self):
...
+        if not self.event_id:
+            return self._take_child_from_website()
```
ecino: This will add the fact that if no children are found on the website, we call the super to still take it from global childpool.
```suggestion
            res = self._take_child_from_website()
            if res:
                return res
```

sms_939/models/sms_child_request.py
```diff
@@ -41,3 +43,47 @@ def send_step1_reminder(self):
...
+        else:
```
ecino:
```suggestion
```

sms_939/models/sms_child_request.py
```diff
@@ -41,3 +43,47 @@ def send_step1_reminder(self):
...
+            return super(SmsRequest, self).reserve_child()
```
ecino:
```suggestion
        return super(SmsRequest, self).reserve_child()
```

sms_939/models/sms_child_request.py
```diff
@@ -41,3 +43,47 @@ def send_step1_reminder(self):
...
+            self.write({
+                'child_id': child.id,
+                'state': 'child_reserved'
+            })
+            child.hold_id.sms_request_id = self.id
+            child.remove_from_wordpress()
+            child.write({
+                'state': 'S'
+            })
+            # Put a new child in event buffer
```
ecino: This shouldn't be done here.
```suggestion
```

sms_939/models/sms_child_request.py
```diff
@@ -41,3 +43,47 @@ def send_step1_reminder(self):
...
+            self.event_id.with_delay().hold_children_for_sms(1)
```
ecino:
```suggestion
```

ecino (changes requested):

ecino (approved):

PR #909, opened 2019-07-02
CO-2644 Muskathlon medical form

Theyiot:
The Muskathlon medical form was deprecated and a new version existing in paper version needed to be created and stored in **Odoo**. This PR solves this problem and provide a French, a German and an English version of the form. The German (mostly) and the English versions need to be checked, since they are not my mother tongues but overall they should be understandable.


ecino:
Will merge with #911

PR #916, opened 2019-07-26
AP-108 App links related work

vwvw:


mobile_app_switzerland/controllers/registration_controller.py
```diff
@@ -20,4 +21,5 @@ def registration_form(self, **kwargs):
         """
         Return registration form
         """
-        return request.redirect('compassionappuk://')
+        hostname = config.get("wordpress_host")
+        return request.redirect('https://' + hostname + '/Login')
```
ecino: I think we wanted to put every mobile page inside the same path. That's what I changed on iOS:
```suggestion
        return request.redirect('https://' + hostname + '/app/Login')
```

ecino (changes requested):

mobile_app_switzerland/controllers/registration_controller.py
vwvw: Good catch!

ecino (approved):

PR #921, opened 2019-08-07
CO-2685

samybucher:
CO-2685
creation of a link between thank you letters and gift. A wizard take the csv file and fill the fields between the objects.
be careful, translation are missing and must be done in the production side

### DIFF
--- a/sponsorship_switzerland/models/load_csv_wizard.py
+++ b/sponsorship_switzerland/models/load_csv_wizard.py
@@ -41,4 +41,4 @@ class LoadCsvWizard(models.TransientModel):
                 if gift and letter:
                     gift.letter_id = letter
                     letter.gift_id = gift
-                    return True
+        return True
--- a/sponsorship_switzerland/models/load_csv_wizard.py
+++ b/sponsorship_switzerland/models/load_csv_wizard.py
@@ -23,7 +23,7 @@ class LoadCsvWizard(models.TransientModel):
     @api.multi
     def link_gift2letters(self):
         for wizard in self:
-            f = StringIO(self.data_csv.decode('base64')
+            f = StringIO(wizard.data_csv.decode('base64')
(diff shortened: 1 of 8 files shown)


sponsorship_switzerland/models/load_csv_wizard.py
```diff
@@ -0,0 +1,43 @@
...
+class LoadCsvWizard(models.TransientModel):
+    _name = 'load.csv.wizard'
+
+    data_csv = fields.Binary('csv file')
+    name_data_csv = fields.Char('File Name')
+
+    @api.multi
+    def link_gift2letters(self):
+        for wizard in self:
+            f = StringIO(self.data_csv.decode('base64')
```
ecino:
```suggestion
            f = StringIO(wizard.data_csv.decode('base64')
```

sponsorship_switzerland/models/load_csv_wizard.py
```diff
@@ -0,0 +1,43 @@
...
+                gift = self.env['sponsorship.gift'].search([
+                    # Dropping the 3 first characters to get the ID
+                    # ex. CH-12345 -> 12345
+                    ('id', '=', row[2][3:])
+                ])
+                if letter.gift_id and gift.letter_id:
+                    continue
+                if gift and letter:
+                    gift.letter_id = letter
+                    letter.gift_id = gift
```
ecino, line 43: At the end of the function, you can `return True` to indicate that everything went smoothly.

ecino (changes requested):

sponsorship_switzerland/models/load_csv_wizard.py
```diff
@@ -0,0 +1,44 @@
...
+                    return True
```
ecino: The return should be outside all loops otherwise you will exit the function at the first iteration of the loop:
```suggestion
        return True
```

PR #947, opened 2019-10-17
Try a more flexible match if the partner can't be found from the bvr …

nbornand:
…reference

### DIFF
--- a/sponsorship_switzerland/models/completion_rules.py
+++ b/sponsorship_switzerland/models/completion_rules.py
@@ -87,7 +87,7 @@ class StatementCompletionRule(models.Model):
         if not partner:
             # Some bvr reference have a wrong number of leading zeros,
             # resulting in the partner reference to be offset.
-            flexible_ref_match = re.search(r'^0{,10}([1-9]\d{1,6})0',
+            flexible_ref_match = re.search(r'^0{,10}([1-9]\d{4,6})0',
                                            ref)
             if flexible_ref_match:
                 flexible_ref = flexible_ref_match.group(1)
--- a/sponsorship_switzerland/models/completion_rules.py
+++ b/sponsorship_switzerland/models/completion_rules.py
@@ -87,7 +87,7 @@ class StatementCompletionRule(models.Model):
         if not partner:
             # Some bvr reference have a wrong number of leading zeros,
             # resulting in the partner reference to be offset.
-            flexible_ref_match = re.search(r'^0{,10}([1-9]\d{6})0',
+            flexible_ref_match = re.search(r'^0{,10}([1-9]\d{1,6})0',
                                            ref)
             if flexible_ref_match:
                 flexible_ref = flexible_ref_match.group(1)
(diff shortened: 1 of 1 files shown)


sponsorship_switzerland/models/completion_rules.py
```diff
@@ -79,9 +80,23 @@ def get_from_partner_ref(self, stmts_vals, st_line):
...
+
+        partner_ref = ref[9:16]
         partner_obj = self.env['res.partner']
-        partner = partner_obj.search(
-            [('ref', '=', str(int(ref[9:16])))])
+        partner = partner_obj.search([('ref', '=', str(int(partner_ref)))])
+        if not partner:
+            # Some bvr reference have a wrong number of leading zeros,
+            # resulting in the partner reference to be offset.
+            flexible_ref_match = re.search(r'^0{,10}([1-9]\d{6})0',
```
ecino: Not sure what you are doing here. Is it correct the following understanding:

- Up to ten 0
- One digit
- Six digits
- One 0

If that's correct I don't understand the middle part with the digits? [1-9] and \d are the same, no?

sponsorship_switzerland/models/completion_rules.py
```diff
@@ -79,9 +80,23 @@ def get_from_partner_ref(self, stmts_vals, st_line):
...
+                                           ref)
+            if flexible_ref_match:
+                flexible_ref = flexible_ref_match.group(1)
+                if int(flexible_ref) != int(partner_ref):
```
ecino, line 94: Not sure why this test is needed? Wouldn't it work if they were equal?

sponsorship_switzerland/models/completion_rules.py
nbornand: [1-9] doesn't include 0. This is useful to separate the leading zeros to the partner id (that cannot start with a 0).

sponsorship_switzerland/models/completion_rules.py
nbornand: If they are equal it means that we have already tries to resolve flexible_ref and don't need to try again.

sponsorship_switzerland/models/completion_rules.py
ecino: There are some smaller refs as well, would it be more risky to change to this?
```suggestion
            flexible_ref_match = re.search(r'^0{,10}([1-9]\d{1,6})0',
```

sponsorship_switzerland/models/completion_rules.py
```diff
@@ -79,9 +80,23 @@ def get_from_partner_ref(self, stmts_vals, st_line):
...
+
+        partner_ref = ref[9:16]
         partner_obj = self.env['res.partner']
-        partner = partner_obj.search(
-            [('ref', '=', str(int(ref[9:16])))])
+        partner = partner_obj.search([('ref', '=', str(int(partner_ref)))])
+        if not partner:
+            # Some bvr reference have a wrong number of leading zeros,
+            # resulting in the partner reference to be offset.
+            flexible_ref_match = re.search(r'^0{,10}([1-9]\d{1,6})0',
```
nbornand:
```suggestion
            flexible_ref_match = re.search(r'^0{,10}([1-9]\d{4,6})0',
```

PR #950, opened 2019-10-21
CO-2772 : shift the ref scan when partner ref is misaligned.

ecino:
I tried to improve further the match flexibility as well with fund types and product, because if partner ref is misaligned, the rest will as well be misaligned. @nbornand Could you please review this change?

### DIFF
--- a/sponsorship_switzerland/models/completion_rules.py
+++ b/sponsorship_switzerland/models/completion_rules.py
@@ -81,7 +81,8 @@ class StatementCompletionRule(models.Model):
             ref = st_line['ref']
         res = {}
 
-        partner_ref = ref[9:16]
+        ref_index = 9  # position of the partner ref inside the BVR number.
+        partner_ref = ref[ref_index:16]
         partner_obj = self.env['res.partner']
         partner = partner_obj.search([('ref', '=', str(int(partner_ref)))])
         if not partner:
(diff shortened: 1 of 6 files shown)


sponsorship_switzerland/models/completion_rules.py
```diff
@@ -236,32 +238,53 @@ def search_partner(criteria):
...
+                ('type', 'like', 'S')
+            ]
+            contract = contract_obj.search(search_criterias + [
+                ('commitment_number', '=', contract_number)])
+            if not contract:
+                # Maybe the ref is misaligned. Most people have a number < 10
+                # so we try to find this digit in the ref and see if we get
+                # lucky.
+                contract_number_match = re.search(
+                    partner.ref + r'0{1,4}(\d)', ref)
```
nbornand, line 283: We are matching a single digit, that can be 0. Is it on purpose? Maybe we should use \d+ or exclude 0 ([1-9]) for the first char.

sponsorship_switzerland/models/completion_rules.py
```diff
@@ -236,32 +238,53 @@ def search_partner(criteria):
...
+        :param stmts_vals: bank statement values
+        :param st_line: current statement line values
+        :param partner: matched partner
+        :param ref_index: starting index at which the partner ref was found
+        :return: dict, boolean: st_line values to update, true if invoice is
                                 created.
         """
         # Read data in english
         res = dict()
+        ref = st_line['ref']
```
nbornand, line 255: nit: could specify that we are talking about the partner red

nbornand (approved):

sponsorship_switzerland/models/completion_rules.py
ecino: Yes it is, as the contract number can be 0 (unfortunately).

PR #952, opened 2019-10-23
Co 2785 cancel invoices on sponsorship termination

nbornand:


sponsorship_switzerland/models/contracts.py
```diff
@@ -380,6 +380,8 @@ def _on_sponsorship_finished(self):
                 sponsorship.correspondent_id.write({
                     'category_id': [(3, sponsor_cat_id),
                                     (4, old_sponsor_cat_id)]})
+            # Deactivate pending invoice lines.
+            sponsorship.invoice_line_ids.mapped('invoice_id').action_cancel()
```
ecino: You should filter on invoice state to cancel only the 'open' invoices.
```suggestion
            sponsorship.invoice_line_ids.mapped('invoice_id').filtered(
                lambda i: i.state == 'open').action_cancel()
```

ecino (changes requested):

ecino (approved):

PR #964, opened 2019-11-04
CO-2791 and CO-2799: Add report 2BVR + fix BVR background

Nooka10:


report_compassion/models/report_bvr_sponsorship.py
```diff
@@ -86,6 +86,24 @@ def render_html(self, docids, data=None):
...
+class TwoBvrSponsorship(models.AbstractModel):
+    _inherit = 'report.report_compassion.bvr_sponsorship'
+    _name = 'report.report_compassion.2bvr_sponsorship'
+
+    def _get_report(self):
+        return self.env['report']._get_report_from_name(
+            'report_compassion.2bvr_sponsorship')
+
+    @api.multi
+    def render_html(self, docids, data=None):
```
ecino: This method is not doing anything else than its parent, I think you can remove it.

report_compassion/models/report_bvr_sponsorship_gift.py
```diff
@@ -61,6 +61,25 @@ def render_html(self, docids, data=None):
...
+class TwoBvrGiftSponsorship(models.AbstractModel):
+    _inherit = 'report.report_compassion.bvr_gift_sponsorship'
+    _name = 'report.report_compassion.2bvr_gift_sponsorship'
+
+    def _get_report(self):
+        return self.env['report']._get_report_from_name(
+            'report_compassion.2bvr_gift_sponsorship')
+
+    @api.multi
+    def render_html(self, docids, data=None):
```
ecino: This method is not doing anything else than its parent, you can remove it.

report_compassion/report/bvr_sponsorship.xml
```diff
@@ -73,23 +87,97 @@
...
+        <t t-set="partner_address" t-value="partner.short_address"/>
+        <t t-set="company_account" t-value="o.get_company_bvr_account()"/>
+
+        <div class="page">
+            <t t-call="report_compassion.bvr_style"/>
+            <t t-foreach="docs.get(o)" t-as="sponsorship">
+                <t t-foreach="sponsorship.get_sponsorship_gift_products()"
+                   t-as="gift">
+                    <t t-call="report_compassion.report_sponsorship_single_slip">
+                        <t t-set="reference"
```
ecino: Missing margin? On other calls you have `<t t-set="margin" t-value="(date_index % 2) * 85"/>` but here you don't define it, is it on purpose?

report_compassion/report/bvr_layout.xml
```diff
@@ -10,8 +10,6 @@
     - o (sponsorship group)
     -->
     <template id="report_sponsorship_single_slip">
-        <t t-set="margin" t-value="position * 0.2"/>
```
ecino, line 13: No margin now, is it correct?

report_compassion/report/bvr_gift.xml
```diff
@@ -22,34 +33,78 @@
...
+        <t t-set="o" t-value="doc.with_context({'lang':doc.partner_id.lang})"/>
         <t t-set="partner" t-value="o.gift_partner_id"/>
         <t t-set="partner_address" t-value="partner.short_address"/>
-        <t t-set="company_account" t-value="o.group_id.get_company_bvr_account()"/>
+        <t t-set="company_account"
+           t-value="o.group_id.get_company_bvr_account()"/>
         <div class="page">
             <t t-call="report_compassion.bvr_style"/>
             <t t-foreach="products" t-as="product">
-                <t t-set="position" t-value="((offset or 0) * product_index) % 3"/>
```
ecino, line 47: I think you should replace by a margin no? Something like this (don't know what 85 stands for) ?
`<t t-set="margin" t-value="(product_index % 3) * 85"/>`

ecino (changes requested):

report_compassion/report/bvr_sponsorship.xml
Nooka10: Margin is used only for the 2BVR. It's not needed for one or 3 BVR.
Do you prefer if I set margin = 0 on those calls?

Nooka10: Margin is used only for the 2BVR. It's not needed for one or 3 BVR.
Do you prefer if I set margin = 0 on those calls?

report_compassion/report/bvr_sponsorship.xml
```diff
@@ -73,19 +88,178 @@
...
+        </t>
+    </template>
+
+    <template id="2bvr_sponsorship">
+        <t t-call="report_compassion.bvr_sponsorship_2bvr"/>
+    </template>
+
+    <!-- 2 BVR page for sponsorship -->
+    <template id="report_bvr_sponsorship_document_2bvr">
+        <t t-set="o" t-# -*- coding: utf-8 -*-
```
ecino: You have something bad here.

ecino (changes requested):
You had an unexpected paste of python code inside an xml, please fix that. Can you as well rebase your branch onto compassion/devel and keep only the relevant commits?

report_compassion/report/bvr_sponsorship.xml
Nooka10: Woups, corrected in 1a0aebaa0d2d0c82377693e465ccad11e411932d

ecino (approved):

PR #975, opened 2019-12-23
CO-2908 1 day birthday reminder changes

quentingigon:


ecino (changes requested):
This is not doing what was asked. Please come so that we can discuss it again.

partner_communication_switzerland/models/contracts.py
```diff
@@ -266,8 +266,33 @@ def _send_reminders_for_birthday_in_1day_or_2months(self):
...
+
+        sponsorships_to_send_reminder = []
+
+        for sponsorship in sponsorships_with_birthday_tomorrow:
+            sponsorship_correspondences = self.env['correspondence'].search([
+                ('sponsorship_id', '=', sponsorship.id),
+                ('direction', '=', "Supporter To Beneficiary"),
+                ('sent_date', '!=', False),
+                ('sent_date', '<=', str(datetime.now())),
+                ('sent_date', '>=', str(datetime.now() - relativedelta(months=-2)))
```
ecino: Replace `sent_date` with `scanned_date`. Only the last one is useful. != False and is less than today are useless.
str(date) is not correct, should be replaced by `fields.Date.to_string()`

partner_communication_switzerland/models/contracts.py
```diff
@@ -266,8 +266,33 @@ def _send_reminders_for_birthday_in_1day_or_2months(self):
         tomorrow = today + relativedelta(days=1)
         sponsorships_with_birthday_tomorrow = \
             self._get_sponsorships_with_child_birthday_on(tomorrow)
+
+        sponsorships_to_send_reminder = []
```
ecino: You are doing the opposite of what is asked! You are sending the reminder to all of those who already wrote or did a gift :(
So I propose to rename this variable to `sponsorships_to_avoid` and make it a recordset instead of a list, because the method `_send_birthday_reminders` asks for a recordset and not a list of ids.
```suggestion
        sponsorships_to_avoid = self.env['recurring.contract']
```

partner_communication_switzerland/models/contracts.py
```diff
@@ -266,8 +266,33 @@ def _send_reminders_for_birthday_in_1day_or_2months(self):
...
+            sponsorship_correspondences = self.env['correspondence'].search([
+                ('sponsorship_id', '=', sponsorship.id),
+                ('direction', '=', "Supporter To Beneficiary"),
+                ('sent_date', '!=', False),
+                ('sent_date', '<=', str(datetime.now())),
+                ('sent_date', '>=', str(datetime.now() - relativedelta(months=-2)))
+            ])
+
+            if sponsorship_correspondences:
+                sponsorships_to_send_reminder.append(sponsorship_correspondences)
```
```suggestion
                sponsorships_to_avoid += sponsorship
```

partner_communication_switzerland/models/contracts.py
```diff
@@ -266,8 +266,33 @@ def _send_reminders_for_birthday_in_1day_or_2months(self):
...
+
+            sponsorship_gifts = self.env['sponsorship.gift'].search([
+                ('sponsorship_id', '=', sponsorship.id),
+            ])
+
+            for gift in sponsorship_gifts:
+                for invoice_line in gift.invoice_line_ids:
+                    if str(datetime.now()) >= invoice_line.last_payment \
+                            >= str(datetime.now() + relativedelta(months=-2)):
+                        sponsorships_to_send_reminder.append(gift.sponsorship_id)
```
ecino: Instead of looping through all the gifts you could directly add the filter in the search like you did with the letters. Use the field `date_partner_paid` in the gift object.

ecino: `sponsorships_to_avoid += sponsorship`

partner_communication_switzerland/models/contracts.py
```diff
@@ -266,8 +266,33 @@ def _send_reminders_for_birthday_in_1day_or_2months(self):
...
+
         self._send_birthday_reminders(
-            sponsorships_with_birthday_tomorrow,
+            sponsorships_to_send_reminder,
```
```suggestion
            sponsorships_with_birthday_tomorrow - sponsorships_to_avoid,
```

ecino (changes requested):

partner_communication_switzerland/models/contracts.py
quentingigon: ..............I must have some problems with this task...

partner_communication_switzerland/models/contracts.py
```diff
@@ -266,8 +266,31 @@ def _send_reminders_for_birthday_in_1day_or_2months(self):
         tomorrow = today + relativedelta(days=1)
         sponsorships_with_birthday_tomorrow = \
             self._get_sponsorships_with_child_birthday_on(tomorrow)
+
+        sponsorships_to_avoid = {}
```
ecino: Empty recordset should be initialised like this:
```suggestion
        sponsorships_to_avoid = self.env['recurring.contract']
```

partner_communication_switzerland/models/contracts.py
```diff
@@ -266,8 +266,31 @@ def _send_reminders_for_birthday_in_1day_or_2months(self):
...
+
+        for sponsorship in sponsorships_with_birthday_tomorrow:
+            sponsorship_correspondences = self.env['correspondence'].search([
+                ('sponsorship_id', '=', sponsorship.id),
+                ('direction', '=', "Supporter To Beneficiary"),
+                ('scanned_date', '>=', fields.Date.to_string(
+                    datetime.now() - relativedelta(months=-2)))
```
ecino: Double negation will lead to an addition!
```suggestion
                    datetime.now() - relativedelta(months=2)))
```

partner_communication_switzerland/models/contracts.py
```diff
@@ -266,8 +266,31 @@ def _send_reminders_for_birthday_in_1day_or_2months(self):
         tomorrow = today + relativedelta(days=1)
         sponsorships_with_birthday_tomorrow = \
             self._get_sponsorships_with_child_birthday_on(tomorrow)
+
+        sponsorships_to_avoid = {}
+
+        for sponsorship in sponsorships_with_birthday_tomorrow:
+            sponsorship_correspondences = self.env['correspondence'].search([
```
ecino: This will be a bit more efficient because it doesn't need to load the records in memory.
```suggestion
            sponsorship_correspondences = self.env['correspondence'].search_count([
```

partner_communication_switzerland/models/contracts.py
```diff
@@ -266,8 +266,31 @@ def _send_reminders_for_birthday_in_1day_or_2months(self):
...
+                    datetime.now() - relativedelta(months=-2)))
+            ])
+
+            if sponsorship_correspondences:
+                sponsorships_to_avoid += sponsorship
+
+            sponsorship_gifts = self.env['sponsorship.gift'].search([
+                ('sponsorship_id', '=', sponsorship.id),
+                ('date_partner_paid', '>=',
+                 fields.Date.to_string(datetime.now() - relativedelta(months=-2)))
```
```suggestion
                 fields.Date.to_string(datetime.now() - relativedelta(months=2)))
```

ecino (changes requested):
Please you should test your changes before pushing. I'm sure this would crash because you are additioning recordsets and dictionaries which are not compatbile.

quentingigon:
I tested and it did not crash... I don't know why

ecino (approved):

PR #984, opened 2020-01-21
CO-2946 applied same changes to welcome email

quentingigon:


quentingigon:
Not yet tested

ecino (changes requested):
This is not good in my opinion. You should try to understand why the method is called only with one record at the time and try to fix that instead, so that when it's called `self` is already the correct recordset (= all sponsorships that require the communication to be sent).

quentingigon:
The jobs are now executed once for all communications of each partner

ecino (approved):

PR #993, opened 2020-01-30
CO-3033 Sms 939 migration v11

quentingigon:


quentingigon:
tests are failing because of GMC for now
Edit: not true, they are bugged

sms_939/models/sms_child_request.py
```diff
@@ -92,16 +90,4 @@ def cancel_request(self):
         if not self.event_id:
             self.child_id.write({'state': 'N'})
             self.child_id.add_to_wordpress()
-        return super(SmsRequest, self).cancel_request()
-
-    @job(default_channel='root.sms_request')
-    @related_action(action='related_action_sms_request')
-    def get_children_from_global_pool_for_website(self, take=1):
```
ecino: You inadvertently removed a method.

sms_939/models/sms_notification.py
```diff
@@ -87,17 +86,19 @@ def run_service(self):
         sms_text = self.text or ''
         if not self.hook_id:
             hooks = self.env['sms.hook'].search([])
+            n1 = "\n"
             sms_answer = _(
                 "Sorry, we could not understand your request. "
-                "Supported services are :\n - %s "
-            ) % "\n- ".join(hooks.mapped('name'))
+                f"Supported services are :\n - {n1 + '- '.join(hooks.mapped('name'))}"
```
ecino, line 92: Why putting \n in a variable?
```suggestion
                f"Supported services are :\n - {'\n-  '.join(hooks.mapped('name'))}"
```

sms_939/wizards/sms_sender_wizard.py
```diff
@@ -63,8 +63,7 @@ def send_sms(self, mobile):
         headers = {}
         username = self.sms_provider_id.username_939
         password = self.sms_provider_id.password_939
-        auth = base64.encodestring('%s:%s' % (username,
-                                              password)).replace('\n', '')
+        auth = f'{username}:{password}'.replace('\n', '')
```
ecino, line 68: This should be base64 encoded!

ecino (changes requested):
I think you mixed up the libraries handling http requests. Why not using the `requests` library which is one that is very simple (and already required by Odoo)? See https://pypi.org/project/requests/

sms_939/models/sms_notification.py
quentingigon: Because expression fragments inside f-strings cannot includes backslashes

sms_939/wizards/sms_sender_wizard.py
ecino: This one is not yet fixed.

ecino (approved):

PR #1006, opened 2020-02-11
CO-2998

josueBulle:
I needed those PR
quentingigon/report_compassion_migration_v11
quentingigon/partner_communication_switzerland_migration_v11
quentingigon/payment_ogone_migration_v11
josueBulle/website_compassion

I also needed a fix for file upload:
https://github.com/CompassionCH/compassion-modules/pull/1288/


eicher31 (approved):
Just fix flake8 error.

josueBulle:
The `default_tickets` function was removed in odoo11. Now I think the default tickets are handled in the `event.type` model. But we use a wizard that (asks for the registration fees and) creates directly  the `event.event`. I did a quick fix but I'm not sure if I should go further and refactor 

https://github.com/odoo/odoo/commit/91617bfc78e3da94be2276fe1c82c4ce47eeb896

PR #1007, opened 2020-02-12
CO-3041 Migrer crm_switzerland

SylvainLosey:


ecino:
Please fix the travis build. It seems that a menu is wrongly inheriting a parent that doesn't exist in v11 in `calendar_event.xml`

eicher31 (approved):

PR #1008, opened 2020-02-12
CO-3040 cleanup_switzerland migration v11

josueBulle:


cleanup_switzerland/models/database_cleanup.py
```diff
@@ -52,7 +51,10 @@ def clean_mail_message(self):
         self.env.cr.execute(
             "DELETE FROM mail_message WHERE model IS NULL"
         )
-        models = self.env['ir.model'].search([]).mapped('model')
+
+        self.env.cr.execute(sql.SQL("SELECT DISTINCT model FROM mail_message"))
+        models = [m[0] for m in self.env.cr.fetchall()]
+
```
josueBulle: The cron was looping on every existing models. Many of them were raising errors because they don't have relations in database (probably the abstract models, almost same behaviour in Odoo10).
So I edited to loop only over models referenced in the mail_message relation.

With an empty database it is really faster but I wonder if it is a efficient with the prod database

cleanup_switzerland/models/database_cleanup.py
```diff
@@ -77,62 +79,69 @@ def clean_mail_message(self):
...
-            "AND res_id IN (SELECT id FROM partner_communication_job WHERE "
-            "state = 'cancel' OR (state = 'done' AND sent_date < %s)"
-            "LIMIT 1000)",
-            [limit_str]
-        )
-        count = self.env.cr.rowcount
-        self.env.cr.commit()  # pylint: disable=invalid-commit
-        _logger.info('{} old partner_communication messages removed'.format(
-            count))
+        if 'gmc.message.pool' in self.env:
```
josueBulle, line 93: I added these if-conditions so I didn't need to add all dependencies in the manifest.

ecino (approved):

PR #1011, opened 2020-02-14
CO-3050  Mysql and sbc switzerland migration v11

quentingigon:


sbc_switzerland/models/advocate_details.py
```diff
@@ -15,7 +14,7 @@
 _logger = logging.getLogger(__name__)
 
 
-class AdvocateDetails(models.Model):
+class AdvocateDetails(models.AbstractModel):
```
ecino: This is wrong, please revert, or maybe it should be GeoModel.

sbc_switzerland/models/correspondence.py
```diff
@@ -189,8 +189,8 @@ def fix_priority(self):
                 'createdat': letter.scanned_date
             })
             text_id = tc.select_one(
-                "SELECT text_id FROM translation WHERE id = %s", [
-                    translation_id]).get('text_id')
+                "SELECT text_id FROM translation "
+                f"WHERE id = {translation_id}").get('text_id')
```
ecino: This is wrong, the SQL queries are an exception where we should not use format, but continue to pass the parameters as exepected by the sql functions. We wisk SQL injections if we do otherwise.

sbc_switzerland/models/correspondence.py
```diff
@@ -213,8 +213,7 @@ def clean_translate(self):
                 "SELECT t.id as id, s.id as status "
                 "FROM translation t join translation_status s on "
                 "s.translation_id = t.id "
-                "WHERE t.letter_odoo_id = %s AND s.status_id = 1", [letter.id]
-            )
+                f"WHERE t.letter_odoo_id = {letter.id} AND s.status_id = 1")
```
ecino: Please revert this one as well.

sbc_switzerland/models/correspondence.py
```diff
@@ -229,11 +228,11 @@ def check_missing_translate(self):
...
             ('state', '=', 'Global Partner translation queue')
         ])
-        logger.info("Found {} letters to transalte.".format(str(len(letters))))
+        logger.info(f"Found {len(letters)} letters to translate.")
         for letter in letters:
             letter_id = tc.select_one(
-                "SELECT id FROM translation WHERE letter_odoo_id = %s",
-                [letter.id]).get('id')
+                "SELECT id FROM translation WHERE"
+                f" letter_odoo_id = {letter.id}").get('id')
```
ecino: Please revert.

sbc_switzerland/models/translate_connector.py
```diff
@@ -106,8 +105,8 @@ def get_lang_id(self, lang_compassion_id):
         """ Returns the language's id in MySQL that has  GP_Libel pointing
          to the iso_code given (returns -1 if not found). """
         res = self.select_one(
-            "SELECT id FROM language WHERE GP_Libel LIKE '{}'"
-            .format(lang_compassion_id.code_iso))
+            "SELECT id FROM language WHERE GP_Libel LIKE "
```
ecino, line 108: Here we should use %s with the parameters of the `select_one` function

sbc_switzerland/models/translate_connector.py
```diff
@@ -161,13 +160,11 @@ def remove_letter(self, text_id):
 
     def remove_from_text(self, text_id):
         """ Delete a text record for the text_id given """
-        self.query("DELETE FROM text WHERE id={}"
-                   .format(text_id))
+        self.query(f"DELETE FROM text WHERE id={text_id}")
```
ecino: Here as well we should use the query parameters with %s

sbc_switzerland/models/translate_connector.py
```diff
@@ -161,13 +160,11 @@ def remove_letter(self, text_id):
...
 
     def remove_translation_with_odoo_id(self, text_id):
-        self.query("DELETE text FROM text INNER JOIN translation ON text.id\
-             = translation.text_id WHERE translation.letter_odoo_id = {}"
-                   .format(text_id))
+        self.query(f"DELETE text FROM text INNER JOIN translation ON text.id\
```
ecino: Here as well we should use the query parameters with %s

sbc_switzerland/models/translate_connector.py
```diff
@@ -201,8 +198,7 @@ def upsert_user(self, partner, create):
 
     def remove_user(self, partner):
         """ Delete a user """
-        return self.query("DELETE FROM user WHERE number={}"
-                          .format(partner.ref))
+        return self.query(f"DELETE FROM user WHERE number={partner.ref}")
```
ecino: Here as well we should use the query parameters with %s

ecino (changes requested):

quentingigon:
There is a bug with the cron : Check translated letters on local translation platform. When opening the mysql connection, it fails with error:
**'Host \'192.168.200.125\' is blocked because of many connection errors; unblock with \'mysqladmin flush-hosts\''**

sbc_switzerland/models/correspondence.py
```diff
@@ -130,7 +130,7 @@ def process_letter(self):
                 if intro_letter in letter.communication_type_ids and not \
                         letter.sponsorship_id.send_introduction_letter:
                     continue
-                if super(Correspondence, letter).process_letter():
+                if super().process_letter():
```
ecino: This should be reverted, because the super was not called on the whole recordset.

sbc_switzerland/models/advocate_details.py
```diff
@@ -48,7 +49,7 @@ def write(self, vals):
         goodbye_config = self.env.ref('sbc_switzerland.translator_goodbye')
         for advocate in self:
             was_translator = translation in advocate.engagement_ids
-            super(AdvocateDetails, advocate).write(vals)
+            super().write(vals)
```
ecino: This should be reverted, the super was not called on the whole recordset.

sbc_switzerland/models/import_letters_history.py
```diff
@@ -61,7 +60,7 @@ def _compute_nber_letters(self):
         for letter in self:
             if letter.manual_import or (
                     letter.state and letter.state != 'draft'):
-                super(ImportLettersHistory, letter)._compute_nber_letters()
+                super()._compute_nber_letters()
```
ecino: This should be reverted.

ecino (changes requested):

ecino (approved):

PR #1016, opened 2020-02-18
CO-3051 Added activities to events

josueBulle:


ecino (approved):

ecino:
After reflextion, can you move modules mis_builder* into compassion-switzerland repository ? It makes more sense there I think.

PR #1031, opened 2020-02-25
CO-3097 Move module settings to general settings

SylvainLosey:


SylvainLosey:
Travis fails because the views depend on CompassionCH/compassion-modules#1297

ecino (changes requested):
Now I have merged the parent Pull Request but we still have an issue here:
`disaster_notify_ids` is not found in the parent view, see the Travis errors and please fix them.

PR #1033, opened 2020-02-26
CO-3099 Fix Postfinance Redirect

SylvainLosey:
In Odoo 11, there are major changes in payment including a [new payment form](https://github.com/odoo/odoo/commit/2df9c22d80be82b4ed7ffcd79f858e46c34e6479#diff-6b3bf59852d51bd845e333022ff9f6ae).

This fix only reuses the v10 `ogone_acquirer_button` instead of the v11 `ogone_form` to render the payment / redirection button. The new Odoo 11 logic doesn't use an html form which also breaks the automatic redirection from [cms_form](https://github.com/CompassionCH/compassion-modules/blob/a9e14965fc281dee2611cecb421e5f7da06f9971/cms_form_compassion/static/src/js/modal_form.js#L17)

### DIFF
--- a/payment_ogone_compassion/__manifest__.py
+++ b/payment_ogone_compassion/__manifest__.py
@@ -36,12 +36,13 @@
     'website': 'http://www.compassion.ch',
     'data': [
         'data/payment_acquirer.xml',
+        'views/payment_ogone_templates.xml'
     ],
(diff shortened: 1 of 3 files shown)


SylvainLosey:
Move to #1035

PR #1034, opened 2020-02-27
Upgrade report compassion

quentingigon:
BVRs are now added as attachment when the communication is sent by email or has not the template "A4 BVR".

A bug was fixed; it should now be possible to choose the order in which we send a communication with send_mode == 'both'

There is still a bug with the body_html of the mail/letter not correctly created

### DIFF
--- a/report_compassion/models/partner_communication.py
+++ b/report_compassion/models/partner_communication.py
@@ -74,7 +74,6 @@ class PartnerCommunication(models.Model):
 
     def _put_bvr_in_attachments(self, bvr, background):
         pdf_download = self._generate_pdf_data(bvr, background)
-        bvr.attachment_ids.unlink()
         return self._create_and_add_attachment(bvr, pdf_download)
 
     def _create_and_add_attachment(self, bvr, datas):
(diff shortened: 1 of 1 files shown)


quentingigon:
The bug with the body_html is resolved but on the PR associed to this one: [HERE](https://github.com/CompassionCH/compassion-modules/pull/1299)

report_compassion/models/partner_communication.py
```diff
@@ -40,11 +41,69 @@ def send(self):
...
+
+        if bvr_to_print:
+            for bvr in bvr_to_print:
+                if bvr.report_id != self.env.ref('report_compassion.report_a4_bvr'):
+                    self._put_bvr_in_attachments(bvr, background=False)
+            return super(PartnerCommunication, bvr_to_print).send()
+
+    def _put_bvr_in_attachments(self, bvr, background):
+        pdf_download = self._generate_pdf_data(bvr, background)
+        bvr.attachment_ids.unlink()
```
ecino: Why are you removing other attachments ? I don't think it's necessary.

ecino (changes requested):

report_compassion/models/partner_communication.py
quentingigon: I did it because I had a bug otherwise but I corrected and refactored that part so it's ok now

ecino (approved):

PR #1036, opened 2020-02-28
CO-3101 Fix send_introduction_letter bug

SylvainLosey:
- Previously if the letter had to be translated, it would not go through the `send_introduction_letter` boolean check as the if statement is inside the translation check. Then it would be sent from `update_translation` without `send_introduction_letter` being checked. To fix this I moved the nested if outside. Couldn't test properly but it's highly likely this was the cause.

- Also found a small problem while trying to reproduce the bug, in the interface if a transfer is set as origin the `send_introduction_letter` field appears and is set to `False`. If the origin is changed the field disappears from the view but is still `False`. Therefore I think we should revert to True when the origin is changed to something that is not a transfer.

### DIFF
--- a/sbc_switzerland/models/correspondence.py
+++ b/sbc_switzerland/models/correspondence.py
@@ -122,14 +122,11 @@ class Correspondence(models.Model):
         intro_letter = self.env.ref(
             'sbc_compassion.correspondence_type_new_sponsor')
         for letter in self:
-            # if letter.original_language_id in \
-            #         letter.supporter_languages_ids or \
-            if (letter.beneficiary_language_ids &
-                    letter.supporter_languages_ids) or \
+            if intro_letter in letter.communication_type_ids and not \
+                    letter.sponsorship_id.send_introduction_letter:
+                continue
+            if (letter.beneficiary_language_ids & letter.supporter_languages_ids) or \
                     letter.has_valid_language:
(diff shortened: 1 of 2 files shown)


ecino:
Please fix the lint checks

sbc_switzerland/models/correspondence.py
```diff
@@ -122,14 +122,11 @@ def process_letter(self):
         intro_letter = self.env.ref(
             'sbc_compassion.correspondence_type_new_sponsor')
         for letter in self:
-            # if letter.original_language_id in \
-            #         letter.supporter_languages_ids or \
-            if (letter.beneficiary_language_ids &
-                    letter.supporter_languages_ids) or \
+            if intro_letter in letter.communication_type_ids and not \
+                    letter.sponsorship_id.send_introduction_letter:
```
quentingigon, line 126: Couldn't we just check only _letter.sponsorship_id.send_introduction_letter_ ?

SylvainLosey: I don't think so, as all letters pass through this function, is is only `intro_letters` where `send_introduction_letter` is false that must be skipped. Otherwise this would avoid sending all letters of sponsorships where `send_introduction_letter` is set to false.

ecino (approved):

PR #1048, opened 2020-03-24
CO-3135 Custom portal layout template correction

quentingigon:
This is not the definitive version but the template is working now

### DIFF
--- a/muskathlon/controllers/main.py
+++ b/muskathlon/controllers/main.py
@@ -126,14 +126,14 @@ class MuskathlonWebsite(EventsController):
         if form_success:
             result = request.redirect("/my/home")
         else:
-            result = request.render("website_portal.portal_my_home", values)
+            result = request.render("muskathlon.custom_portal_my_home", values)
         return self._form_redirect(result, full_page=True)
 
     @route(["/my/api"], type="http", auth="user", website=True)
     def save_ambassador_picture(self, **post):
         user = request.env.user
         partner = user.partner_id
-        return_view = "website_portal.portal_my_home"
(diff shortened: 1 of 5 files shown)


quentingigon:
The problem I have is that I have no example of what it must look like, as the my_home page is not working on devel and on erp.compassion. There is something but the formatting is bad, and I don't know if I must just make it working or perfect it

muskathlon/templates/muskathlon_my_home.xml
```diff
@@ -51,7 +51,7 @@
                     <div id="modal_flight_details" class="modal fade cms_modal_form" tabindex="-1" role="dialog"
                          aria-labelledby="donation_modal_label" aria-hidden="true">
                         <!-- TODO: When upgrading to Bootstrap 4.2 or higher, use class="modal-dialog modal-xl" -->
```
ecino: You can remove the comment now :)

muskathlon/controllers/main.py
```diff
@@ -47,7 +47,7 @@ def my_muskathlon_details(self, registration, **kwargs):
         )
         return request.render("muskathlon.my_details", {"reports": reports})
 
-    @route(["/my", "/my/home"], type="http", auth="user", website=True)
+    @route(["/my", "/my/home"], type="http", auth="user", website=True, cors="*")
```
ecino: Just out of curiosity, what is `cors` used for?

quentingigon: I added it because there are CORS error when loading the page and I found other places where this is done. But it doesn't do anything, so I will remove it

ecino (approved):

PR #1050, opened 2020-03-25
CO-3161 Crowdfunding_compassion module creation

quentingigon:
Created module that will be use for the crowd-funding platform.
It adds a website for the crowd-fundings and a new Bootstrap theme (in the form of a new module)


crowdfunding_compassion/data/crowdfunding_website.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Monthly Cron for Accounting Entries -->
```
ecino: Unpurposeful comment

theme_crowdfunding/static/src/scss/bootstrap.scss
```diff
@@ -0,0 +1,44 @@
+/*!
```
ecino: Do we really need bootstrap and all these scss ?? I think it should be included in Odoo...

ecino (changes requested):
- In the module __init__ file, you should import your models.
- Does the theme have to be in a separate module in order to work ? I would prefer to put everything in the same module `crowdfunding_compassion` if that's possible. If not possible, then you should add dependency to the module `theme_crowdfunding` in `crowdfunding_compassion`'s manifest
- Please refer to this documentation : https://www.odoo.com/documentation/12.0/howtos/themes.html
  I believe that you integrated wrongly the whole bootstrap framework instead of just defining the components you need. Make usage of snippets where useful.

quentingigon:
- From what I read, a theme must be a standalone module but I haven't tried differently. 
- The models are part of the other PR [HERE](https://github.com/CompassionCH/compassion-switzerland/pull/1052), so the import is made there
- You are right for the bootstrap files, it is an oversight and I will correct it while making the snippets

ecino:
Yes, maybe we need a separate module for the theme. You can look as well how Sylvain did for the compassion theme, he did a separate module which has not so much css files in it.

ecino (approved):

PR #1053, opened 2020-03-31
Crowdfunding_compassion module creation

quentingigon:

PR #1055, opened 2020-04-06
CO-3164 [Crowdfunding] projects list page

quentingigon:


quentingigon:
I will rebase once [this PR](https://github.com/CompassionCH/compassion-switzerland/pull/1053) is merged

quentingigon:
I think this page is almost done, except for a bit of css

ecino:
> I will rebase once [this PR](https://github.com/CompassionCH/compassion-switzerland/pull/1053) is merged

It is merged. I see a lot of commits. Is it possible to rebase now and squash some commits ? Is it ready for review or which PR is best to start with ?

crowdfunding_compassion/__manifest__.py
```diff
@@ -34,19 +34,23 @@
     "license": "AGPL-3",
     "website": "http://www.compassion.ch",
     "depends": [
-        "cms_form_compassion",          # compassion-modules
-        "theme_crowdfunding",           # compassion-switzerland
-        "website_event_compassion"      # compassion-switzerland
+        "cms_form_compassion",  # compassion-modules
+        "theme_crowdfunding",  # compassion-switzerland
+        "website_event_compassion",  # compassion-switzerland
```
ecino: I think the module should be able to exist independently from the events website, why do you need this dependence ? I would suggest to replace by `crm_compassion` which is the events but without the website frontend.

crowdfunding_compassion/controllers/main.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class CrowdFundingWebsite(EventsController):
+
+    @route(["/my_account"], type="http", auth="user", website=True)
+    def my_account(self, form_id=None, **kw):
+        """ Inject data for forms. """
+
+        result = request.render("crowdfunding_compassion.myaccount_view_template")
+        return self._form_redirect(result, full_page=True)
```
ecino: The `__form_redirect` redirect method is defined in module `cms_form_compassion` and is used for payment forms. Not sure you need it here, could you just return result instead?

crowdfunding_compassion/controllers/project_controller.py
```diff
@@ -0,0 +1,70 @@
+from datetime import datetime
+
+from babel.dates import format_timedelta
+
+from odoo.http import request, route, Controller
+
+
+class ProjectController(Controller):
+    # Demo route used to display demo project
+    @route("/project/", auth="public", website=True)
```
ecino: Maybe use a less generic route than `/project` for a demo page. `/project/demo` would be better. But this will make the page available to everyone so not sure it's good to keep it like that. Maybe add a TODO comment to remove this route after we finish with the developement?

crowdfunding_compassion/controllers/project_controller.py
```diff
@@ -0,0 +1,70 @@
...
+        demo_project = request.env.ref(
+            "crowdfunding_compassion.demo_project_crowdfunding"
+        ).sudo()
+        print(f"Demo project ID: {demo_project.id}")
+        return request.render(
+            "crowdfunding_compassion.project_page",
+            self._prepare_project_values(demo_project, **kwargs),
+        )
+
+    @route(["/project/<int:project_id>"], auth="public", website=True)
```
ecino: All the routes will start with `/together`. The platform will be called "Together". So here it should be `/together/project/<int...>`

crowdfunding_compassion/controllers/projects_controller.py
```diff
@@ -0,0 +1,42 @@
...
+#    Releasing children from poverty in Jesus' name
+#    @author: Quentin Gigon
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+
+from odoo.http import request, route
+from odoo.addons.website_event_compassion.controllers.events_controller import (
+    EventsController,
```
ecino: I think you can inherit a more generic controller

crowdfunding_compassion/controllers/projects_controller.py
```diff
@@ -0,0 +1,42 @@
...
+)
+
+
+class ProjectsController(EventsController):
+
+    @route('/projects',
```
ecino: `/together/projects`

ecino (changes requested):

crowdfunding_compassion/controllers/main.py
quentingigon: I thought we had to use it and it causes me some problems so I am very happy to hear you say that and I will replace it

crowdfunding_compassion/controllers/project_controller.py
quentingigon: this route is not used anymore, I will just remove it

ecino (approved):

PR #1056, opened 2020-04-07
CO-3163 [Crowdfunding] Project page

SylvainLosey:
-  With a standard Odoo install, change domain of the Crowfunding website to `crowdfunding.localhost:8069` for testing. The demo project page is then available on `crowdfunding.localhost:8069/project/demo`

- Update `theme_crowdfunding` if the css doesn't work (General settings > Website > Crowdfunding > Select Theme > Update)

**TODO**

- [x] Implement project x time left
- [x] Implement share project part (waiting on assets from Olivier)
- [x] Implement fund icon and name selector (waiting on assets from Olivier)


ecino:
I merged #1055 . I believed your PR depends on it and has duplicate commits. You should rebase now on the 12.0 branch and keep only the relevant commits for your changes.

SylvainLosey:
Rebase is done and the PR is ready for review.

Some notes:
- Some files, for example most of the CSS, have already been merged in #1055 and don't appear in this PR
- For some reason while I specify `"sponsorship_switzerland.product_template_fund_csp"` as fund in the demo project another seemingly random fund is used instead. As most funds don't have the newly added field `crowdfunding_impact_text`, this part appears empty in the page but works when completing the field from the backend
- I'm wondering if participants can participate in multiple projects ? If so some adjustments have to be made to the models (ex we would need multiple `number_sponsorships_goal` depending upon the project)
- It will be easier to add some parts (ex fund card) when the task implementing them is merged. I will create a jira task with the elements that have to be added later.

crowdfunding_compassion/controllers/project_controller.py
```diff
@@ -7,7 +7,20 @@
 
 class ProjectController(Controller):
 
-    @route(["/together/project/<int:project_id>"], auth="public", website=True)
```
ecino, line 10: Please restore this route path /together prefix for all routes in this module.

crowdfunding_compassion/data/crowdfunding_website.xml
```diff
@@ -3,7 +3,7 @@
     <data>
         <record id="crowdfunding_website" model="website">
             <field name="name">Crowdfunding Website</field>
-            <field name="domain">localhost:8069</field>
+            <field name="domain">crowdfunding.localhost:12069</field>
```
ecino: Is this needed? Why another port?

crowdfunding_compassion/data/demo.xml
```diff
@@ -53,7 +53,7 @@
         <field name="twitter_url">https://twitter.com</field>
         <field name="instagram_url">https://instagram.com</field>
         <field name="personal_web_page_url">https://example.com</field>
-        <field name="product_id" ref="sponsorship_switzerland.product_template_fund_aid"/>
+        <field name="product_id" ref="sponsorship_switzerland.product_template_fund_csp"/>
```
ecino: The problem you have here in my opinion is that `product_id` references a `product.product` record while `product_template_fund_csp` is a `product.template` record. That's why you get a wrong model. You can replace this like that to fix it:
```suggestion
        <field name="product_id" eval="ref('sponsorship_switzerland.product_template_fund_csp').product_variant_ids[:1].id"/>
```

ecino (changes requested):

ecino:
> * I'm wondering if participants can participate in multiple projects ? If so some adjustments have to be made to the models (ex we would need multiple `number_sponsorships_goal` depending upon the project)

To answer this question : A crowdfunding.participant will only be linked to one crowdfunding.project. If a same person wants to participate to multiple projects, it will create new crowdfunding.participant records for each participation, and link all of them to the same res.partner object.

crowdfunding_compassion/controllers/project_controller.py
SylvainLosey: I added the `/together` prefix, but is it necessary if together/crowdfunding is in the subdomain/domain ?

crowdfunding_compassion/data/crowdfunding_website.xml
SylvainLosey: The sudomain (or another domain) is necessary for the multi website to correctly separate themes. The port however is just my peronal install (docker), I set it back to 8069.

crowdfunding_compassion/data/demo.xml
SylvainLosey: Yes now I have the good product thank you! However the `crowdfunding_impact_text` I added in `products.xml` doesn't seem to be registered on the related `product.product`.

SylvainLosey:
> To answer this question : A crowdfunding.participant will only be linked to one crowdfunding.project. If a same person wants to participate to multiple projects, it will create new crowdfunding.participant records for each participation, and link all of them to the same res.partner object.

Oh I see, all good then.

crowdfunding_compassion/data/crowdfunding_website.xml
ecino: Thanks for the explanation

crowdfunding_compassion/controllers/project_controller.py
ecino: What will be the final route then? Is it `crowdfunding.localhost:8069/together/project` or still `localhost:8069/together/project`? If the answer is the first, then simply change `crowdfunding` subdomain name by `together`and you can remove `/together` from the route. But could the routes conflict with other routes that would have the same path but in another subdomain?

theme_crowdfunding/static/src/style/custom.scss
```diff
@@ -41,6 +41,8 @@ $font-size-sm: $font-size-base * 0.75;
 $font-family-sans-serif: "Montserrat", sans-serif;
 $body-color: #333;
 
+// TODO: Don't apply css globally
```
ecino, line 44: What does it mean? Did you enter a task in Jira for that? Otherwise we risk to forget it.

theme_crowdfunding/views/assets.xml
```diff
@@ -2,6 +2,7 @@
 <odoo>
     <template id="theme_crowdfunding_assets" name="Assets for crowdfunding website" inherit_id="website.assets_frontend">
         <xpath expr="link[last()]" position="after">
+            <link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" integrity="sha384-wvfXpqpZZVQGK6TAh5PVlGOfQNHSoD2xbE+QkPxCAFlNEevoEH3Sl0sibVcOQVnN" crossorigin="anonymous"/>
```
ecino: Why do you need fontawesome css, isn't it included with Odoo?

crowdfunding_compassion/data/demo.xml
ecino: Yes it might add the impact text in the template but not in the related products that are already existing. For that to work you would need a migration script. But I think we can leave it like that and configure manually our products once we will put it into production.

crowdfunding_compassion/controllers/project_controller.py
SylvainLosey: I think the best would be `together.compassion.ch/project`. I haven't tested but yes the same path on different websites/subdomains will probably conflict. If we need to have identical paths we can choose in the controller which template/page to render depending on the currently visited website/subdomain. Or use different routes as we don't have that many.

theme_crowdfunding/static/src/style/custom.scss
SylvainLosey: You're right, I will add a Jira task.

theme_crowdfunding/views/assets.xml
SylvainLosey: I didn't know they were, removed it.

ecino (approved):

PR #1062, opened 2020-04-17
CO-3165 [Crowdfunding] Create homepage

SylvainLosey:
- Had a weird bug with the migration, will need to check if the migration works on devel

**TODO**

- [x] Find how to display correctly as front page (for now `/home`, `/` does not work)
- [x] Link up activated funds


ecino:
- Did you have contact with Olivier for missing assets or should I take care of it? If so please, tell me what is missing.
- All routes should start with `/together` (name of the platform). The home page should not be home because it's maybe reserved. `/together/home` should work. What are your issues there? I think we will just create a domain that will redirect to the correct page.

SylvainLosey:
- Yes I have informed Olivier some time ago, but I will remind him he may have forgotten.
- From what I understand multi websites in Odoo are differentiated by their domain, so I thought the website would be either a subdomain (ex together.compassion.ch) or an other domain (ex together.ch). So I want the homepage to appear at the root of the domain and I just need to check how to do that correctly

ecino:
Great, thank you. FYI I merged #1055 and it looks that you have conflicts with it that you should fix. Please inform me when review is requested.

SylvainLosey:
The PR is rebased and ready for review.

- I wasn't able to set the homepage to the root of the domain. By overriding the controller we override homepages of other websites so this doesn't work. By creating a `website.page` linking to the template on route `/` we can't access the data passed by the controller (projects, funds etc). Just using `/` as route of the controller doesn't work either. As a result I left the page on `/home` for now.
- The migration linking the existing toilets fund to the xml will need to be tested on devel
- I will create a jira task for what is left to do at the end of the sprint

crowdfunding_compassion/controllers/homepage_controller.py
```diff
@@ -0,0 +1,54 @@
+from datetime import datetime
+
+from odoo.http import request, route, Controller
+
+
+class HomepageController(Controller):
+    @route("/home", auth="public", website=True)
```
ecino: The module should have all routes starting with /together as a prefix for the platform.
```suggestion
    @route("/together/home", auth="public", website=True)
```

crowdfunding_compassion/migrations/12.0.1.0.1/pre-migration.py
```diff
@@ -0,0 +1,21 @@
+from openupgradelib import openupgrade
```
ecino: This migration is for the module `sponsorship_switzerland`. I think your bug comes from that because sponsorship_switzerland will be updated before crowdfunding_compassion and therefore your migration is not working here. Move it to sponsorship_switzerland so that it executes before it reads the xml data files of the module sponsorship_Switzerland.

ecino (changes requested):

crowdfunding_compassion/migrations/12.0.1.0.1/pre-migration.py
SylvainLosey: I moved it to sponsorship_switzerland but I still get the same error. Maybe this is somehow related to the docker installation I have.

ecino:
I think the branch has conflicts with your previous PR. Please check this.

SylvainLosey:
PR is rebased :)

ecino (approved):

PR #1068, opened 2020-04-28
12.0 crowdfunding fixes

ecino:
Various template fixes including streamlined header and footer accross the pages.


website_event_compassion/models/event_compassion.py
```diff
@@ -86,8 +85,6 @@ def _compute_event_type(self):
                 event.type = "presentation"
             elif event.event_type_id == meeting:
                 event.type = "meeting"
-            elif event.event_type_id == crowdfunding:
-                event.type = "crowdfunding"
```
quentingigon, line 90: This was done to fix the bug occurring when we create a project from the crowdfunding website (when we create a crm.event.compassion). Is it not needed anymore ?

quentingigon (commented):
Everything looks good except the change in website_event will give an error.

SylvainLosey (approved):
Looks good to me

website_event_compassion/models/event_compassion.py
ecino: The way it was done is wrong because website_event_compassion is not aware of crowdfunding and should not put this type because it will crash if the crowdfunding module is not installed. I will test again.

quentingigon: Oh ok, that's right. So we need to implement that change in the crowdfunding module ?

ecino: I fixed it in my last commit.

quentingigon (approved):

PR #1069, opened 2020-04-30
[Crowdfunding] CO-3169 Donation process

SylvainLosey:
- Wasn't able to properly test successful payments and email confirmation
- Postfinance might need to be preselected in the donation form logic as default payment provider
- When testing if a psycopg2 error occurs when submitting the payment form, replace `                                "account_id": product.property_account_income_id.id` by `"account_id": 169` in donation_form. Cause is accounts aren't available in empty databases.
- What I did to add the terms and conditions and alternative donation inside the rendered form is very hacky, open to suggestions to do it in another way.


ecino:
I merged some changes of Quentin that brought conflicts with your code now. Please have a look.

crowdfunding_compassion/__manifest__.py
```diff
@@ -34,16 +34,17 @@
...
-        "event",                    # odoo base modules
+        "crm_compassion",  # compassion-modules
+        "event",  # odoo base modules
     ],
     "data": [
         "data/crowdfunding_website.xml",
         "data/crowdfunding_event_type.xml",
-        "data/project_confirmation_email.xml",
+        "data/email_templates.xml",
+        "data/demo.xml",
```
ecino: Please remove demo file, it's already loaded in the demo section.

crowdfunding_compassion/__manifest__.py
```diff
@@ -55,13 +56,11 @@
         "views/product_template_view.xml",
         "views/projects_list_page.xml",
         "views/project_creation_page.xml",
-        "views/crowdfunding_components.xml",
```
ecino: Why removing this view?

crowdfunding_compassion/__manifest__.py
```diff
@@ -55,13 +56,11 @@
...
         "views/project_page.xml",
-        "views/progress_bar.xml",
```
ecino, line 64: Why removing this view?

crowdfunding_compassion/data/demo.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ecino: Please merge this into the demo file inside the demo folder.

crowdfunding_compassion/views/project_donation_form_page.xml
```diff
@@ -0,0 +1,154 @@
...
+                    <div class="container mb-5">
+                        <h3 class="blue uppercase">Your billing information</h3>
+                        <div t-attf-class="{{form.form_wrapper_css_klass}}">
+                            <t t-raw="form.form_render()"/>
+                        </div>
+                    </div>
+                </section>
+            </t>
+
+            <script>
```
ecino: Shouldn't the script be placed in a separate js file and loaded properly with assets_frontend? You could also make benefit from jQuery library.

crowdfunding_compassion/views/project_donation_form_page.xml
```diff
@@ -0,0 +1,154 @@
...
+                // Register on each radio change
+                radios.forEach(radio => {radio.addEventListener('click', clickHandler);});
+
+                // Exectute on page load
+                clickHandler();
+            </script>
+        </template>
+
+        <!-- Add Terms and Conditions to donation form -->
+        <template id="donation_form_gtc" inherit_id="cms_form.base_form">
```
ecino: Didn't you see that we have a widget for terms and condition fields in CMS forms ? You could have used that or did you face issues with it?

crowdfunding_compassion/views/project_donation_page.xml
```diff
@@ -0,0 +1,128 @@
...
+                        </div>
+                        <a id="url" href="">
+                            <button id="submit" type="submit" class="btn btn-primary btn-lg mb-5 float-right uppercase">Next step</button>
+                        </a>
+                    </div>
+                </section>
+
+                <!-- This function disables the next button if there isn't 2 radios checked
+                (fund/sponsor and participant) and adjusts the link for next step -->
+                <script>
```
ecino: Could we move this as well into a separate js file?

theme_crowdfunding/static/src/style/custom.scss
```diff
@@ -574,151 +583,95 @@ footer {
...
-$mute: #DFE3E4;
-
-.progress {
-  list-style: none;
-  margin: 0;
-  padding: 0;
-  display: table;
-  table-layout: fixed;
+// Fix cms.form checkboxes
+.field-boolean {
```
ecino, line 631: Won't this change the styles for all forms in all pages/themes ? Maybe it's risky and should use more specific selectors or is it only applied when the crowdfunding_theme is active (would be good)?

ecino (changes requested):

ecino:
Please check the conflicts and also the comments I made. Maybe moving the JS part can be done afterwards.

crowdfunding_compassion/__manifest__.py
SylvainLosey: It was both in crowdfuning_components and here, so I thought this was an old file. I can split it again, but I don't know if we should group reused components in one file or split them by function ?

crowdfunding_compassion/views/project_donation_form_page.xml
SylvainLosey: Sure, I will look into it along with my next task so that we can merge this PR.

crowdfunding_compassion/views/project_donation_form_page.xml
SylvainLosey: I saw this afterwards, but I didn't test it. I can switch to it if you want ? It might need some refactoring as then the anonymous checkbox won't have the same styles / alignment

crowdfunding_compassion/views/project_donation_page.xml
SylvainLosey: Will do

theme_crowdfunding/static/src/style/custom.scss
SylvainLosey: It was my assumption that it was only applied when the theme is active, otherwise the theme wouldn't be working. From what I tested this is the case.

crowdfunding_compassion/__manifest__.py
ecino: It's ok then, I thought it was a mistake and didn't know the view is already in another file.

PR #1071, opened 2020-05-04
CO-3207 12.0 finish project page

quentingigon:


crowdfunding_compassion/controllers/project_controller.py
```diff
@@ -57,7 +57,7 @@ def _prepare_project_values(self, project, **kwargs):
                 "color": "grey",
                 "text": f"{int(donation.price_unit / donation.product_id.list_price)} "
                 f"{donation.product_id.crowdfunding_impact_text_passive}",
-                "image": "TODO: fund icon selector",
+                "image": donation.product_id.product_tmpl_id.image_medium,
```
ecino:
```suggestion
                "image": donation.product_id.image_medium,
```

crowdfunding_compassion/controllers/project_controller.py
```diff
@@ -69,7 +69,12 @@ def _prepare_project_values(self, project, **kwargs):
         # Chronological list of sponsorships and fund donations for impact display
         impact = sorted(sponsorships + donations, key=lambda x: x["date"])
 
-        return {"project": project, "impact": impact}
+        fund = request.env['product.product'].sudo().search([
+            ('activate_for_crowdfunding', '=', True),
+            ("name", "like", project.product_id.product_tmpl_id.name)
+        ])
```
ecino: I made the same suggestion in other PR.
```suggestion
        fund = project.product_id
```
You could directly access it from the template like that without needing to pass it as a separate variable.

crowdfunding_compassion/templates/project_page.xml
```diff
@@ -87,22 +87,24 @@
                                     <t t-esc="project.description"/>
                                 </p>
                                 <t t-if="project.presentation_video">
-                                    <iframe class="my-3" width="100%" height="400px" t-att-src="project.presentation_video" frameborder="0"/>
+                                    <iframe class="my-3" width="100%" height="400px" t-attf-src="#{project.presentation_video}" frameborder="0"/>
```
ecino: This would have also worked with previous version. t-att takes variable and  interpret directly as python code. t-attf writes as a string and uses special notation #{} to inject python. So the previous version was even better in my opinion.

theme_crowdfunding/views/assets.xml
```diff
@@ -4,6 +4,10 @@
         <xpath expr="link[last()]" position="after">
             <link href="/theme_crowdfunding/static/src/style/custom.scss" rel="stylesheet" type="text/scss"/>
             <link href="/theme_crowdfunding/static/src/style/stepper.scss" rel="stylesheet" type="text/scss"/>
+            <link href="https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css"
+                  rel="stylesheet"
+                  integrity="sha384-wvfXpqpZZVQGK6TAh5PVlGOfQNHSoD2xbE+QkPxCAFlNEevoEH3Sl0sibVcOQVnN"
+                  crossorigin="anonymous"/>
```
ecino: Don't inject fontawesome as it is already loaded into Odoo.

ecino (changes requested):

crowdfunding_compassion/controllers/project_controller.py
quentingigon: I will change the template later

crowdfunding_compassion/templates/project_page.xml
quentingigon: ohh ok I didn't know, I will change it back

ecino (approved):

PR #1074, opened 2020-05-06
Final text content and general improvements

SylvainLosey:
In preparation of the translations task, I added the definitive texts translated in english as well as translation functions in python code. In doing so I also fixed and improved several areas of the website.


crowdfunding_compassion/models/crowdfunding_project.py
```diff
@@ -77,27 +80,31 @@ def create(self, vals):
...
+    #     """Add the project owner to the participant list. """
+    #     for project in self:
+    #         if project.project_owner_id not in project.participant_ids.mapped(
+    #             "partner_id"
+    #         ):
+    #             participant = {
+    #                 "partner_id": project.project_owner_id.id,
+    #                 "project_id": project.id,
+    #             }
+    #             project.write({"participant_ids": [(0, 0, participant)]})
```
ecino: I think rather than commenting this, you should remove at the other place if it was done manually elswhere to create the owner participant.

ecino (changes requested):

crowdfunding_compassion/models/crowdfunding_project.py
SylvainLosey: Ok

ecino (approved):

PR #1075, opened 2020-05-07
CO-3203 Translations of crowdfunding platform

SylvainLosey:
- Did my best to find the appropriate text, but not easy to find each english keyword out of context in a document with everything in another order. I translated what I couldn't find and parts that weren't translated, but it will need to be checked by the translators and corrected.
- Will rebase when #1074 is merged


ecino:
You can rebase now

ecino (approved):

PR #1076, opened 2020-05-08
12.0 registration steps

ecino:


crowdfunding_compassion/forms/project_creation_form.py
```diff
@@ -8,16 +8,47 @@
...
+
+    def form_check_permission(self, raise_exception=True):
+        # no need for this
+        pass
+
+    def wiz_configure_steps(self):
+        return {
+            1: {"form_model": "cms.form.crowdfunding.project.step1"},
+            2: {"form_model": "cms.form.crowdfunding.project.step1"},
+            3: {"form_model": "cms.form.crowdfunding.project.step1"},
```
SylvainLosey: Not sure but should't these be cms.form.crowdfunding.project.step2 and cms.form.crowdfunding.project.step3 ?

ecino: Oh yes :)

SylvainLosey:
Looks good to me otherwise :)

crowdfunding_compassion/controllers/projects_controller.py
```diff
@@ -2,43 +2,80 @@
...
 class ProjectsController(Controller, FormControllerMixin):
     @route("/projects", auth="public", website=True)
     def get_projects_list(self, **kwargs):
-        project_obj = request.env["crowdfunding.project"]
+        project_obj = request.env["crowdfunding.project"].sudo()
 
         # TODO connect pagination to backend -> CO-3213
         return request.render(
             "crowdfunding_compassion.project_list_page",
             {"projects": project_obj.sudo().get_active_projects()},
```
quentingigon, line 25: this sudo is not needed anymore I think

ecino: you are right

crowdfunding_compassion/models/crowdfunding_participant.py
```diff
@@ -18,9 +18,9 @@ class CrowdfundingParticipant(models.Model):
         "res.partner", string="Partner", required=True, index=True, ondelete="cascade"
     )
     personal_motivation = fields.Text()
-    product_number_goal = fields.Integer(default=1)
+    product_number_goal = fields.Integer(default=0)
```
quentingigon, line 21: This will make an error when displaying a goal_progress progress bar, as product_number_goal is used as "diviseur" and if == 0, it will crash

crowdfunding_compassion/models/crowdfunding_participant.py
```diff
@@ -18,9 +18,9 @@ class CrowdfundingParticipant(models.Model):
...
     product_number_reached = fields.Integer(compute="_compute_product_number_reached")
-    number_sponsorships_goal = fields.Integer(default=1)
+    number_sponsorships_goal = fields.Integer(default=0)
```
quentingigon, line 23: same here

quentingigon (commented):
Looking good to me, except my few comments

crowdfunding_compassion/models/crowdfunding_participant.py
ecino: Then we should fix the progress bar because projects can have 0 product goal and only focus on sponsorships.

quentingigon: @SylvainLosey should maybe refactor it

SylvainLosey: > This will make an error when displaying a goal_progress progress bar, as product_number_goal is used as "diviseur" and if == 0, it will crash

If there is only sponsorships as goal and `product_number_goal` is 0, then `product_id` (the fund) would be null and the progress bar wouldn't be displayed.

SylvainLosey: If there is a situation where `product_id` is set but `product_number_goal` is 0, then we should just modify `goal_progress` a bit:

```<t t-set="progress_percentage" t-value="goal_reached / goal_objective * 100 if goal_objective else 0"/>```

ecino: I have made the corrections.

PR #1083, opened 2020-05-18
Participant page

quentingigon:
I think all the functionalities are there, the task is not finished yet as I need to test it a bit more and the design may need to be changed but I create the PR to see if I go in the right direction


quentingigon:
It should be more or less finished

ecino:
> It should be more or less finished

What does this mean?

quentingigon:
The functionalities are there but maybe the design should change

crowdfunding_compassion/controllers/project_controller.py
```diff
@@ -22,8 +22,50 @@ def project_page(self, project, **kwargs):
             self._prepare_project_values(project.sudo(), **kwargs),
         )
 
-    # TODO: test when we can create data for a project
+    @route(["/participant/<int:project_id>/<int:participant_id>"],
```
ecino: The participant is unique for each project, because if the same person registers for multiple projects it will create a new participant record. Therefore I propose to only use participant in the route and use model for more user friendly route (names instead of ids only).
```suggestion
    @route(["/participant/<model('crowdfunding.participant'):participant>/"],
```

crowdfunding_compassion/controllers/project_controller.py
```diff
@@ -22,8 +22,50 @@ def project_page(self, project, **kwargs):
...
+           type="http",
+           auth="user",
+           website=True)
+    def participant(self, project_id=0, participant_id=0):
```
ecino:
```suggestion
    def participant(self, participant=None, **kwargs):
```

crowdfunding_compassion/controllers/project_controller.py
```diff
@@ -22,8 +22,50 @@ def project_page(self, project, **kwargs):
...
+        participant = request.env['crowdfunding.participant'].sudo().search([
+            ("id", "=", participant_id)
+        ])
+        project = request.env['crowdfunding.project'].sudo().search([
+            ("id", "=", project_id)
+        ])
```
ecino:
```suggestion
        project = participant.project_id
```

crowdfunding_compassion/templates/participant_page.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
ecino: Couldn't you reuse the project template and inject different values in variables ? This way we would avoid too much code duplication and the need to change both templates the day we want to make some design changes.

ecino (changes requested):

crowdfunding_compassion/templates/participant_page.xml
quentingigon: yes

ecino:
Please check pylint errors + the failing tests

ecino:
I think this commit is no more relevant and should be removed : https://github.com/CompassionCH/compassion-switzerland/pull/1083/commits/5aeea109d8c33847cda9ce87eb5b68135899e968

ecino:
Please check travis, your code is broken: https://travis-ci.org/github/CompassionCH/compassion-switzerland/builds/692064595

quentingigon:
I dont have the results of Travis on this page

crowdfunding_compassion/__manifest__.py
```diff
@@ -63,7 +63,7 @@
         "templates/project_creation_page.xml",
         "templates/project_donation_form_page.xml",
         "templates/project_donation_page.xml",
-        "templates/project_page.xml",
```
ecino, line 66: I think you removed project_page by mistake? or you should delete the file project_page.xml if it's no more needed.

ecino (changes requested):

crowdfunding_compassion/__manifest__.py
quentingigon: I forgot to remove it when I merged the 2 presentation pages (project and participant) in the same file

ecino (approved):

PR #1087, opened 2020-05-19
CO-3226 new large image field in product.product

quentingigon:


crowdfunding_compassion/models/product_template.py
```diff
@@ -23,3 +23,6 @@ class ProductTemplate(models.Model):
     fund_selector_post_description = fields.Char(
         translate=True, help="Ex: children"
     )
+    image_large = fields.Binary(
+        "Large image", help="Image for header"
```
ecino: To avoid storing binaries in database
```suggestion
        "Large image", help="Image for header", attachment=True
```

ecino (changes requested):

ecino (approved):

PR #1088, opened 2020-05-19
CO-3232 participant profile image saved in partner

quentingigon:


crowdfunding_compassion/forms/project_creation_form.py
```diff
@@ -288,6 +288,7 @@ def form_after_create_or_update(self, values, extra_values):
                 "crowdfunding_compassion.config_project_confirmation").sudo()
             participant = self.main_object.sudo().participant_ids
             partner = self.main_object.sudo().project_owner_id
+        partner.sudo().write({"image": extra_values.get('partner_image')})
```
ecino: This should be done only if `extra_values.get('partner_image')` has a value. Otherwise we risk to remove an existing picture (in case he doesn't upload any image).

ecino (changes requested):

crowdfunding_compassion/forms/project_creation_form.py
quentingigon: very true, I'll change it

ecino (approved):

PR #1089, opened 2020-05-20
CO-3190 - Migrate JS of account_reconcile_compassion to V12

SylvainLosey:
With this migration, `account_reconcile_create_invoice` is merged into `account_reconcile_compassion`. This is done for code simplification, and because creating a custom reconcile view requires overloading and rewriting one function (`_renderCreate()`) of the original module, which is easier to do in one place.

The JS has been migrated and all previous functions it implemented should be functional. However as there was heavy changes from the previous version, there might be further changes necessary on the python / xml side. I don't fully understand how the module is supposed to be used correctly, so this will need to be looked at with/by someone who has better knowledge of the module.


eicher31:
There are some pylint issue, could you think you can fix it?

ecino:
I think this is the last error:
```
account_reconcile_compassion/models/account_reconcile_model.py:6: [R7980(consider-merging-classes-inherited), AccountReconcileModel] Consider merging classes inherited to "account.reconcile.model" from models/statement_operation.py:16.
```

SylvainLosey:
Travis is happy :)

PR #1091, opened 2020-05-26
Fix better zip bug

quentingigon:
Half of fix for better_zip problems, other half is [HERE](https://github.com/CompassionCH/compassion-modules/pull/1336)


quentingigon:
As it's done for now, it will make a difference between for example 1400, Yverdon and 1400, Yverdon-les-Bains

PR #1092, opened 2020-05-26
CO-3235 Fix Welcome active sent multiple times

SylvainLosey:
Hypothesis: some error / timeout occurs when sending the communication, and as such the sponsorships never get welcome_active_letter_sent set to True. If this is the case, by setting to true first, we ensure that it is not sent multiple times, but some communications may not be sent.


eicher31 (approved):
Error flake8:
partner_communication_switzerland/models/contracts.py:364:5: E303 too many blank lines (2)

PR #1095, opened 2020-05-29
CO-3245 Notify participant on new donation

SylvainLosey:


crowdfunding_compassion/forms/donation_form.py
```diff
@@ -103,3 +103,15 @@ def form_after_create_or_update(self, values, extra_values):
         self.amount = extra_values.get("amount")
         self.is_anonymous = extra_values.get("anonymous_donation")
         super().form_after_create_or_update(values, extra_values)
+
+        # Send email to inform participant of new donation
+        participant = self.main_object.sudo()
```
ecino: I think this is too early in the process to send the confirmation, because the person might cancel or abandon the payment. Therefore you should identify the invoice and only send the confirmation when the invoice is paid. This could be done in the confirmation route : https://github.com/CompassionCH/compassion-switzerland/blob/12.0/crowdfunding_compassion/controllers/donation_controller.py#L61

ecino (changes requested):

ecino (approved):

PR #1103, opened 2020-06-12
CO-3001 Implement Auto Reconciliation

SylvainLosey:
- Created a new method instead of overriding _apply_rules() to have cleaner code, and avoid unnecessary computation when it is called during bank statement import, which will be most of the time.
- Added a button to the bank statement view to launch the function but might not be necessary

### DIFF
--- a/account_reconcile_compassion/models/bank_statement_line.py
+++ b/account_reconcile_compassion/models/bank_statement_line.py
@@ -63,7 +63,7 @@ class BankStatementLine(models.Model):
 
     @api.multi
     @job(default_channel="root.bank_reconciliation")
-    def process_reconciliation(
+    def _process_reconciliation(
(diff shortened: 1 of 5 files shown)


account_reconcile_compassion/models/bank_statement_line.py
```diff
@@ -155,6 +62,7 @@ def get_statement_line_for_reconciliation_widget(self):
     ##########################################################################
 
     @api.multi
+    @job(default_channel="root.bank_reconciliation")
     def process_reconciliation(
```
ecino: You could rename it to `_process_reconciliation` and add another method `process_reconciliation` that calls `self.with_delay()._process_reconciliation()` in order to force this method to always run in a separate thread.

ecino (approved):

PR #1107, opened 2020-06-18
CO-3277 recreate payment.icon records in xml

wvmrtn:


ecino:
Thanks for finding what the issue is. However it's better like this to force reload some records that were flagged noupdate (otherwise you will reload it each time you update the module). See my commit https://github.com/CompassionCH/compassion-switzerland/commit/e0bd59e6efa7bc2ce4301fa71a4bb24f8c848fed

PR #1114, opened 2020-06-23
CO-3264 campaign duplicate fix

jo-bloiz-ornot:


jo-bloiz-ornot:
linked pull request (mentioned above)

PR #1123, opened 2020-06-26
CO-3234 fix fund_id uniqueness in product_template

wvmrtn:
Note that if there are existing records for the model in the database, then any field values have to be unique, or _sql_constraints attribute of the model will fail and OpenERP will do nothing to enforce the constraint for new records. This if the case for fund_id = 0, which is already non-unique for product.tempate.

Implemented _sql_contraints attribute nonetheless but added api.constrains decorator on fund_id.

### DIFF
--- a/partner_compassion/models/product.py
+++ b/partner_compassion/models/product.py
@@ -7,13 +7,33 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
-from odoo import models, fields
+from odoo import api, models, fields
+from odoo.exceptions import ValidationError
+
+
+def check_fund_id(record):
+    products = record.env[record._name].search([
+        ('fund_id', '!=', 0),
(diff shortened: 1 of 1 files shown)


partner_compassion/models/product.py
```diff
@@ -7,13 +7,33 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
-from odoo import models, fields
+from odoo import api, models, fields
+from odoo.exceptions import ValidationError
+
+
+def check_fund_id(record):
```
ecino, line 14: Why doing an external public method instead of coding it directly inside the `_check_fund_id` method?

wvmrtn: Wanted to make this method available for other methods in the future, such as product.product below in the script.

ecino (approved):

PR #1127, opened 2020-06-29
CO-3206 fix name inversion when creating linked parnter

wvmrtn:
This bug is fixed by changing the key's value of partner_names_order in Settings>Technical>Parameters>System Parameters from first_last to last_first_comma.


ecino (changes requested):
We don't want to change this. We want firstname first and lastname second.

ecino:
This doesn't fix the issue.

PR #1130, opened 2020-06-30
CO-3261_search_bank_address

jo-bloiz-ornot:
Migration from v10 and add a fix in bonus


jo-bloiz-ornot:
edit manifest with pull request https://github.com/CompassionCH/compassion-switzerland/pull/1157

PR #1153, opened 2020-07-08
CO-3214 replaced language of gift BVR from partner's to correspondent's

wvmrtn:
When printing gift payment slips, switched language from the partner's to the correspondent's.

### DIFF
--- a/report_compassion/models/contract.py
+++ b/report_compassion/models/contract.py
@@ -51,7 +51,7 @@ class Contract(models.Model):
     @api.multi
     def get_gift_communication(self, product):
         self.ensure_one()
-        lang = self.correspondent_id.lang
+        lang = self.mapped(self.send_gifts_to).lang
         child = self.child_id.with_context(lang=lang)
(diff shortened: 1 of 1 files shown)


Theyiot (commented):
LGTM

report_compassion/models/contract.py
```diff
@@ -51,7 +51,7 @@ def _compute_amount_due(self):
     @api.multi
     def get_gift_communication(self, product):
         self.ensure_one()
-        lang = self.partner_id.lang
+        lang = self.correspondent_id.lang
```
ecino: In fact the gift could be either sent to the payer or the correspondent. There is a field in the contract determining this and you could use it to use the correct partner.
```suggestion
        lang = self.mapped(self.send_gifts_to).lang
```

ecino (changes requested):

Theyiot:
Required modifications were applied, the _PR_ is now ready.

PR #1159, opened 2020-07-14
mise à jour des champs bvr

quillasp:


Theyiot (approved):
Changes look good to me but the qualification to verify that the fields are now right is the most important. @Quillasp Please in the future insert the name of the task in the _PR_ name to ease linkage between _Jira_ and _Github_

ecino (approved):
I will merge manually to edit the commit name

PR #1160, opened 2020-07-14
CO-3340 Add language selector to Muskathlon

SylvainLosey:
I'm not sure if there isn't an italian translation of muskathlon events or if there is another problem with the translation


Theyiot (approved):
The modifications in the code look good to me.

ecino (approved):

PR #1163, opened 2020-07-15
CO-3359 fix the image encoding

quillasp:


Theyiot:
From my tests the solution works fine and can be integrated.

Theyiot (approved):
The code looks good to me and from my tests, the fix works and can be integrated.

ecino (approved):

PR #1166, opened 2020-07-16
changements dans à la ligne 405

quillasp:
voilà voilà

### DIFF
--- a/partner_communication_switzerland/models/contracts.py
+++ b/partner_communication_switzerland/models/contracts.py
@@ -402,8 +402,14 @@ class RecurringContract(models.Model):
             ]
         )
         if to_send:
-            to_send.send_communication(welcome, both=True).send()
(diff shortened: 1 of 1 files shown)


partner_communication_switzerland/models/contracts.py
```diff
@@ -402,8 +402,14 @@ def _send_welcome_active_letters_for_activated_sponsorships(self):
             ]
         )
         if to_send:
```
Theyiot: This line can be removed, if no records complies the search criterion, there will be no record to iterate on and we won't even enter the `for` loop.

Theyiot (changes requested):
There is one requested change. Moreover, try to use more explicit commits and _PR_ names and leave the description empty instead of filling it with information that is not revelant. The rest of the code looks good to me.
@ecino @eicher31 The modification already appears on branch 12.0, the changes should be reverted and applied from this _PR_.

PR #1170, opened 2020-07-20
CO-3357 Crowdfunding UTM update

SylvainLosey:
Two part PR, other part is here: https://github.com/CompassionCH/compassion-modules/pull/1398

- Add a crowdfunding origin for each project
- Add UTM data to account.invoice.line
- Remove redudant crowdfunding_project_id field from account.invoice.line

### DIFF
--- a/crowdfunding_compassion/forms/donation_form.py
+++ b/crowdfunding_compassion/forms/donation_form.py
@@ -89,7 +89,6 @@ class CrowdfundingDonationForm(models.AbstractModel):
                                 "source_id": participant.source_id.id,
                                 "medium_id": medium.id,
                                 "campaign_id": project.campaign_id.id,
-                                "account_analytic_id": event.analytic_id.id,
                             },
                         )
                     ],
--- a/crowdfunding_compassion/forms/donation_form.py
+++ b/crowdfunding_compassion/forms/donation_form.py
@@ -60,6 +60,7 @@ class CrowdfundingDonationForm(models.AbstractModel):
         event = project.event_id
         product = project.product_id
         name = f"[{project.name}] Donation for {participant.partner_id.name}"
+        medium = self.env.ref("crowdfunding_compassion.utm_medium_crowdfunding")
         partner = self.partner_id.sudo()
         partner.write({"state": "active"})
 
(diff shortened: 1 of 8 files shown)


crowdfunding_compassion/forms/donation_form.py
```diff
@@ -84,8 +85,11 @@ def generate_invoice(self):
                                 "account_analytic_id": event.analytic_id.id,
                                 "user_id": participant.partner_id.id,
                                 "crowdfunding_participant_id": participant.id,
-                                "crowdfunding_project_id": project.id,
                                 "is_anonymous": self.is_anonymous,
+                                "source_id": participant.source_id.id,
+                                "medium_id": medium.id,
+                                "campaign_id": project.campaign_id.id,
+                                "account_analytic_id": event.analytic_id.id,
```
ecino: This line is already at 85

crowdfunding_compassion/models/crowdfunding_project.py
```diff
@@ -68,7 +68,7 @@ class CrowdfundingProject(models.Model):
         compute="_compute_sponsorships"
     )
     invoice_line_ids = fields.One2many(
-        "account.invoice.line", "crowdfunding_project_id", string="Donations"
+        "account.invoice.line", "crowdfunding_participant_id", string="Donations"
```
ecino: This won't work because the inverse field `crowdfunding_participant_id` points to a participant and not a project. It will therefore link all donations of the participant that has the same id as the project which is wrong! The only possible way to have this relation if the inverse Man2one doesn't exist anymore is to do a compute field.

crowdfunding_compassion/models/account_invoice_line.py
```diff
@@ -10,7 +10,4 @@ class AccountInvoiceLine(models.Model):
     crowdfunding_participant_id = fields.Many2one(
         "crowdfunding.participant", "Crowdfunding participant"
     )
-    crowdfunding_project_id = fields.Many2one(
```
ecino, line 13: Could you also add a `onchange` method in the invoice line so that when someone changes the participant from the form view of an invoice line, it automatically sets the other fields `source_id`, `medium_id` and `campaign_id` ?

ecino (changes requested):

crowdfunding_compassion/models/account_invoice_line.py
SylvainLosey: Yes I added this, except for medium_id which is always crowdfunding

ecino (approved):

PR #1171, opened 2020-07-20
CO 3342 12.0 fix bug on payment mode change

wvmrtn:
Fixed an undesirable situation where the payment method is modified and saved then an error is raised. In this case, the status of the related contracts are modified but not reverted, even though the payment mode is effectively reverted.

### DIFF
--- a/sponsorship_switzerland/models/contract_group.py
+++ b/sponsorship_switzerland/models/contract_group.py
@@ -73,14 +73,10 @@ class ContractGroup(models.Model):
             payment_name = payment_mode.name
             contracts |= self.mapped("contract_ids")
             for group in self:
-                old_term = group.payment_mode_id.name
                 if "LSV" in payment_name or "Postfinance" in payment_name:
-                    group.contract_ids.contract_waiting_mandate()
                     # LSV/DD Contracts need no reference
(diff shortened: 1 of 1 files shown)


Theyiot (approved):
LGTM

sponsorship_switzerland/models/contract_group.py
```diff
@@ -101,6 +97,26 @@ def write(self, vals):
...
+        # payment method did not successfully change.
+        if "payment_mode_id" in vals:
+            payment_mode = (
+                self.env["account.payment.mode"]
+                    .with_context(lang="en_US")
+                    .browse(vals["payment_mode_id"])
+            )
+            payment_name = payment_mode.name
+            for group in self:
+                old_term = group.payment_mode_id.name
```
ecino, line 112: This cannot be right, because you moved the check after the `super` call. However when calling `super`, the new values are written. So the old_term will always be equal to the new term here. All your block should be done before calling `super`

ecino (changes requested):

sponsorship_switzerland/models/contract_group.py
wvmrtn: Corrected with commit 529ffdb

ecino (approved):

PR #1172, opened 2020-07-20
CO-3380 Crowdfunding add default profile pics

SylvainLosey:
The uncovered case (No profile pic, Family title) currently doesn't display any profile pic.


crowdfunding_compassion/models/crowdfunding_participant.py
```diff
@@ -39,6 +39,7 @@ class CrowdfundingParticipant(models.Model):
     instagram_url = fields.Char(string="Instagram link")
     personal_web_page_url = fields.Char(string="Personal web page")
     profile_photo = fields.Binary(related="partner_id.image")
```
Theyiot, line 41: With the new implementation, this field is only used to compute the value of the variable `profile_photo_url`. Both fields could be merged to avoid storing duplicated information.

Theyiot (changes requested):
Apart from one change request, the rest of the _PR_ looks good to me and can be merged once the modifications done, if considered necessary.

crowdfunding_compassion/models/crowdfunding_participant.py
ecino: A related field is only a shortcut and does not store the information twice, so that's not really an issue here, unless it would have the attribute `store=True`

crowdfunding_compassion/models/crowdfunding_participant.py
```diff
@@ -84,3 +85,18 @@ def _compute_sponsorships(self):
     def _compute_number_sponsorships_reached(self):
         for participant in self:
             participant.number_sponsorships_reached = len(participant.sponsorship_ids)
+
+    @api.multi
+    def _compute_profile_photo_url(self):
+        domain = self.env['website'].get_current_website()._get_http_domain()
```
ecino, line 91: Did you test this in private tab? I wonder if an unregistered user has security access to `get_current_website` otherwise you should add `sudo()` Unless the sudo is already handled upper in the code each time the participant are accessed from the website. In that case you could as well remove the `sudo` calls below.

crowdfunding_compassion/models/crowdfunding_participant.py
Theyiot: Oops, I missed this `related=...` option, my bad !

crowdfunding_compassion/models/crowdfunding_participant.py
```diff
@@ -84,3 +85,18 @@ def _compute_sponsorships(self):
...
+            if participant.profile_photo:
+                path = f"web/content/crowdfunding.participant/{participant.id}/profile_photo"
+            elif participant.sudo().partner_id.title.name == "Mister":
+                path = "crowdfunding_compassion/static/src/img/guy.png"
+            elif participant.sudo().partner_id.title.name == "Madam":
+                path = "crowdfunding_compassion/static/src/img/lady.png"
+            else:
+                path = None
+
+            participant.profile_photo_url = f"{domain}/{path}" if path else path
```
ecino, line 102:
```suggestion
            participant.profile_photo_url = f"{domain}/{path}"
```

crowdfunding_compassion/models/crowdfunding_participant.py
```diff
@@ -39,6 +39,7 @@ class CrowdfundingParticipant(models.Model):
     instagram_url = fields.Char(string="Instagram link")
     personal_web_page_url = fields.Char(string="Personal web page")
     profile_photo = fields.Binary(related="partner_id.image")
+    profile_photo_url = fields.Binary(compute="_compute_profile_photo_url")
```
Theyiot: This field should be of type `fields.Char`, since the _URL_ is stored as a string and not the binary image directly.

Theyiot (changes requested):

crowdfunding_compassion/models/crowdfunding_participant.py
SylvainLosey: Oops

crowdfunding_compassion/models/crowdfunding_participant.py
SylvainLosey: Yes it's working so they seem to have access too. The sudo below are required however, I had rights error messages without them, logged in and logged out.

SylvainLosey: Yes it's working so they seem to have access too. The sudo below are required however, I had rights error messages without them, logged in and logged out.

ecino (approved):

PR #1173, opened 2020-07-20
CO-3376 reduced size of titles for smaller screens

wvmrtn:
Reduced size of titles display-1 and display-4 for screens of width smaller than 425px.


Theyiot (approved):
The changes are in accordance with the requirements, LGTM.

PR #1178, opened 2020-07-23
CO-3329 Add project type filter to crowdfunding

SylvainLosey:


Theyiot (approved):
Nice new functionality and well integrated, good job !

crowdfunding_compassion/models/crowdfunding_project.py
```diff
@@ -220,35 +220,34 @@ def validate(self):
...
 
-        return active_projects + finished_projects
+        projects = active_projects + finished_projects
+
+        filters = [
+            ("state", "!=", "draft"),
+            ("deadline", ">=", datetime(year, 1, 1)) if year else None,
+            ("deadline", "<=", datetime(year, 12, 31)) if year else None,
+            ("type", "=", type) if type else None
+        ]
```
ecino: Why not directly searching in the beginning (line 225) according to these filters and then only processing those projects?

crowdfunding_compassion/models/crowdfunding_project.py
```diff
@@ -220,35 +220,34 @@ def validate(self):
...
+
+        return projects.search(list(filter(None, filters)))
```
ecino: I think the `search` method acts on all records, this line would be equivalent of `self.env['crowdfunding.project'].search([..])`, meaning that it could return projects that were not present in the records at the beginning. That's why I suggest to start by filtering before doing the join of finished and active projects.

ecino (changes requested):

crowdfunding_compassion/models/crowdfunding_project.py
SylvainLosey: Oh I see, that's kind of an odd behavior. I updated the function and applied the filters in the same search() call as for the dates.

ecino (approved):

PR #1183, opened 2020-07-24
CO-3392 Fix domain for mail activities, set employee field for partners

Theyiot:
The domain of the `mail.activities` has been set to consider only the employees of _CompassionCH_. There is a field `employee` inside `res.partner` but that is false for every single record of the database. A migration is also included in this _PR_ to set this field correctly for the employees of the _NGO_.

### DIFF
--- a/partner_compassion/models/mail_activity.py
+++ b/partner_compassion/models/mail_activity.py
@@ -1,0 +1,20 @@
+##############################################################################
+#
+#    Copyright (C) 2020 Compassion CH (http://www.compassion.ch)
+#    @author: Théo Nikles <theo.nikles@gmail.com>
+#
+#    The licence is in the file __manifest__.py
+#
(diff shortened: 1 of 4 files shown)


partner_compassion/models/mail_activity.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class MailActivity(models.Model):
+    _inherit = "mail.activity"
+
+    # Override domain to consider only internal users
+    user_id = fields.Many2one(
+        "res.users", "Assigned to",
+        default=lambda self: self.env.user,
+        index=True, required=True,
```
ecino: When overriding you can only replace the attributes you want and don't specify the others that were already defined in the parent.
```suggestion
```

ecino (approved):

PR #1186, opened 2020-07-24
CO-3390 FIX A4 + BVR printed twice

SylvainLosey:
- Previously creating a BVR attachment was skipped for self.env.ref("report_compassion.report_a4_bvr") but it must be skipped for A4 OMR BVR too, which is another report but shares the same `report_name`: "report_compassion.a4_bvr". Therefore I changed the check to this field, hopefully this isn't too broad.
- Skip printing when getting the pdf data for sending by email


Theyiot (approved):
LGTM

ecino (approved):

PR #1189, opened 2020-07-27
CO 3385 12.0 add security group criminal records

wvmrtn:
- Added a security group for viewing/downloading criminal records
- Added Date field representing date at which criminal record has been uploaded
- Refactored _compute and _inverse of criminal records in event.registration and res.partner to take into account the date at which the criminal recorded has been uploaded and to save the attachment with the partner name
- Created an action rule to notify event responsible when an attendee's criminal record is expiring


partner_compassion/models/partner_compassion.py
```diff
@@ -11,6 +11,9 @@
 import tempfile
 import uuid
 
+from odoo.addons.website_event_compassion.models.\
```
ecino: The problème is that `partner_compassion` doesn't depend on module `website_event_compassion`, it is in fact a parent module! Therefore you cannot call methods of this module from there. You should move the method here and do the opposite in `website_event_compassion`. You can as well make the method public since it's needed outside of the module.

partner_compassion/models/partner_compassion.py
```diff
@@ -215,6 +223,71 @@ def _compute_survey_input_count(self):
...
+                                "completed_task_ids": [
+                                    (
+                                        4,
+                                        registration.env.ref(
+                                            "website_event_compassion.task_criminal"
+                                        ).id,
+                                    ),
+                                ]
+                            }
+                        )
```
ecino: You cannot access event registrations from this module. You should inherit the method inside `website_event_compassion` to add this logic where you want.

partner_compassion/models/partner_compassion.py
```diff
@@ -389,6 +462,33 @@ def action_view_partner_invoices(self):
...
+        """
+        # search for event responsible(s) of all events attended by partner
+        for partner in self:
+            registrations = self.env['event.registration'].sudo().search([
+                ("partner_id", "=", partner.id)
+            ])
+            for registration in registrations:
+                # get responsible
+                event = registration.event_id
+                responsible = event.user_id
```
ecino: You shouldn't take the responsible of the event. Some people may also have a criminal record but not any event registrations, so this should work regardless of events. You should add a setting to define the user to notify inside the Compassion Settings (by inheriting res.config.settings like it's done in other modules) and use this setting to select the user.

ecino (changes requested):

partner_compassion/models/partner_compassion.py
wvmrtn: Corrected with commit 592de88

partner_compassion/models/partner_compassion.py
wvmrtn: Corrected with commit ca8ec5c

partner_compassion/models/partner_compassion.py
wvmrtn: Corrected in commit 643bc23

PR #1192, opened 2020-07-29
Update project cards CSS

SylvainLosey:
Several little things were bugging me in the crowdfunding project cards. I took the opportunity to do some small changes to the CSS and hopefully have a better presentation. Of course the changes are a proposition and can be discussed.

- Have a uniform height for the title part and more generally for all sections so that the cards elements are aligned.
- Truncate the title and description through CSS to occupy all available space
- Leave all vertical space to the title and move collective/individual to the second row as most titles are long and could use more space
- Update the layout to leave whitespace between the description / bars and the buttons instead of spacing evenly between the element which leads to them being misaligned
- Small margin and elements size updates

Before

Screenshot from 2020-07-29 12-21-09

After
Screenshot from 2020-07-29 12-18-36

### DIFF
--- a/crowdfunding_compassion/templates/crowdfunding_components.xml
+++ b/crowdfunding_compassion/templates/crowdfunding_components.xml
@@ -18,7 +18,7 @@
                     </span>
                 </div>
             </div>
-            <div class="goal-progress__bar mb-2">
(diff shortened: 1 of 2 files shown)


ecino (approved):
Far better, thanks :+1:

PR #1193, opened 2020-07-29
CO 3401 12.0 limit crowdfunding title length

wvmrtn:
Limit Crowdfunding project name to 100 (arbitrary) characters.


ecino:
Wouldn't be sufficient to add the attribute on the field `name` of the crowdfunding.project `size=100` ?

wvmrtn:
> Wouldn't be sufficient to add the attribute on the field `name` of the crowdfunding.project `size=100` ?

The field 'name' actually belongs to a model in the oca addons. Tried overwriting this field in crowdfunding.project with 'size=100' but this change broke various things, namely project creation on the together website and the crowdfunding project records.

Theyiot (approved):
The code looks good to me. It seems to handle properly quite a few border cases. For the records, the four "tolerated" keys are _backspace_, _delete_, _left_ and _right_. I think that those four are enough to have the expected behaviour.

PR #1194, opened 2020-07-30
CO-3403 wrong family salutation

quillasp:
Now, a family should be properly greeted. There might be some people registered as family even though they're just one person, so the result might be unexpected

### DIFF
--- a/partner_communication_switzerland/models/res_partner.py
+++ b/partner_communication_switzerland/models/res_partner.py
@@ -42,9 +42,7 @@ class ResPartner(models.Model):
     @api.multi
     def _compute_salutation(self):
         # Special salutation for companies
-        company = self.filtered(
-            lambda p: not (p.title and p.firstname and not p.is_company)
-        )
+        company = self.filtered('is_company')
         for p in company:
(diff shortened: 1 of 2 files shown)


partner_communication_switzerland/models/res_partner.py
```diff
@@ -55,6 +53,19 @@ def _compute_salutation(self):
...
+            title_salutation = (
+                partner.env["ir.advanced.translation"]
+                    .get("salutation", female=title.gender == "F",
+                         plural=title.plural)
+                    .title()
+            )
+            title_name = title.name
+            partner.salutation = (
+                    title_salutation + " " + title_name + " " + partner.lastname
+            )
```
Theyiot: This does not change much, but the assignation to a variable with a name that have the same length and is used only once can be removed.
```suggestion
            partner.salutation = (
                    title_salutation + " " + title.name + " " + partner.lastname
            )
```

Theyiot (changes requested):
Except one little detail, the rest of the code looks good to me.

ecino (approved):

PR #1200, opened 2020-08-03
CO-3261_search_bank_address

jo-bloiz-ornot:
Migration from v10 and add a fix in bonus

### DIFF
--- a/partner_compassion/models/partner_compassion.py
+++ b/partner_compassion/models/partner_compassion.py
@@ -232,7 +232,8 @@ class ResPartner(models.Model):
             "physical": "paper",
         }
         for partner in self:
-            partner.thankyou_letter = thankyou_mapping[partner.thankyou_preference]
+            partner.thankyou_letter = \
+                thankyou_mapping[partner.thankyou_preference]
 
     ##########################################################################
     #                              ORM METHODS                               #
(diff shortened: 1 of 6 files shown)


partner_compassion/models/partner_compassion.py
```diff
@@ -26,6 +26,16 @@
...
+THANKYOU_MAPPING = {
+    "none": "no",
+    "auto_digital": "default",
+    "auto_digital_only": "only_email",
+    "auto_physical": "paper",
+    "digital": "default",
+    "digital_only": "only_email",
+    "physical": "paper",
+}
+
```
eicher31: This should not be in this pr...

partner_compassion/models/partner_compassion.py
```diff
@@ -222,17 +232,9 @@ def _compute_thankyou_letter(self):
...
-            "auto_digital_only": "only_email",
-            "auto_physical": "paper",
-            "digital": "default",
-            "digital_only": "only_email",
-            "physical": "paper",
-        }
         for partner in self:
-            partner.thankyou_letter = thankyou_mapping[partner.thankyou_preference]
+            partner.thankyou_letter = \
+                THANKYOU_MAPPING[partner.thankyou_preference]
```
eicher31: same here

PR #1203, opened 2020-08-06
CO-3427 Fixing crowdfunding broken links in specific cases

Theyiot:
There used to be a problem if a user entered a link that did not begin either with `http://` or `https://`. The server would interpret the link as a route and append the address of the server before, e.g. `https://erp.compassion.ch/www.mysite.com`. The `http://` is now forced in such cases.
Also, there was a bug that would make the server return an _error 500_ if the video link did not contain a proper hostname and a path. This is also handle in this _PR_.

### DIFF
--- a/crowdfunding_compassion/forms/project_creation_form.py
+++ b/crowdfunding_compassion/forms/project_creation_form.py
@@ -112,7 +112,7 @@ class ProjectCreationWizard(models.AbstractModel):
         if values.get(key):
             url = values[key]
             if not url.startswith("http"):
-                return "http://" + url
+                return "https://" + url
         return None
 
 
--- a/crowdfunding_compassion/forms/project_creation_form.py
+++ b/crowdfunding_compassion/forms/project_creation_form.py
@@ -13,6 +13,8 @@ from base64 import b64encode, b64decode
 from PIL import Image
 from io import BytesIO
 
+import urllib.parse as urlparse
+
 from odoo import models, fields, _
 from odoo.tools import file_open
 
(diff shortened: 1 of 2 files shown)


crowdfunding_compassion/forms/project_creation_form.py
```diff
@@ -105,6 +107,14 @@ def compress(image):
             return new_image if bytes_len > new_bytes_len else old_image
         return old_image
 
+    def sanitized_url(self, values, key):
+        # Prevent the URL to be considered as a route if necessary
+        if values.get(key):
+            url = values[key]
+            if not url.startswith("http"):
+                return "http://" + url
```
ecino: Should we use https by default? I think it's widely used nowadays...
```suggestion
                return "https://" + url
```

Theyiot: It depends on what behaviour we want to achieve. Using `http://` the secured websites will automatically be redirected on `https://` and it will work normally for websites that are not secured. If we use by default `https://`, it will work normally for secured website but people will have a warning that they are going on an unsecured website if it is `http://` instead.
But `https://` should be the standard nowadays so maybe the second option is the best. Let me know what you think.

ecino: I think the second option still encourages the good practices and we better warn or users if they leave a secure environment.

Theyiot: Perfect, I added the modification

ecino (approved):

PR #1206, opened 2020-08-10
CO-3418 LSV and DD payments in progress are not closed with sponsorship

Theyiot:


ecino (changes requested):
This is a good change when sponsorship is finished but I think we miss the case where we edit the amount of the sponsorship. It won't go through this code and the same problem happens. Could you check as well that the second case works?

ecino (approved):

PR #1208, opened 2020-08-11
CO-3423 Donations are now displayed in my_account

Theyiot:


crowdfunding_compassion/controllers/main.py
```diff
@@ -24,8 +24,12 @@ def my_account(self, form_id=None, **kw):
                 ("project_id.project_owner_id", "!=", partner.id),
             ]
         )
-        donations = participations.mapped("invoice_line_ids").filtered(
-            lambda l: l.state == "paid")
+        donations = request.env["account.invoice.line"].sudo().search(
+            [
+                ("crowdfunding_participant_id.partner_id", "=", partner.id),
```
ecino, line 29: You can directly filter in the search instead of doing 2 pass.
```suggestion
                ("crowdfunding_participant_id.partner_id", "=", partner.id),
                ("state", "=", "paid"),
```

crowdfunding_compassion/controllers/main.py
```diff
@@ -24,8 +24,12 @@ def my_account(self, form_id=None, **kw):
...
+            ]
+        )
+        donations = donations.filtered(lambda l: l.state == "paid")
```
ecino:
```suggestion
```

ecino (changes requested):

crowdfunding_compassion/controllers/main.py
Theyiot: Indeed, I missed that and this will be way better !

crowdfunding_compassion/controllers/main.py
Theyiot: This is fixed by the next commit.

ecino (approved):

PR #1209, opened 2020-08-11
Co 3280 Remove claim_type use since it's not usefull anymore

nfluckiger:


ecino (approved):
Make sure you make your PR to devel branch. Otherwise it's fine :+1:

PR #1211, opened 2020-08-13
CO-3206 FIX partner name inversions

ecino:
Fix by forcing lastname and firstname when copying a partner
and removing field name from default values to avoid
a wrong inverse computation


Theyiot (approved):
The modifications look good to me, this _PR_ can go to qualification.

PR #1217, opened 2020-08-17
CO-3422 Applying requested changes

Theyiot:
In parallel to: https://github.com/CompassionCH/compassion-modules/pull/1427.

### DIFF
--- a/partner_communication_switzerland/data/sponsorship_communications_cron.xml
+++ b/partner_communication_switzerland/data/sponsorship_communications_cron.xml
@@ -39,9 +39,10 @@
             <field name="config_id" ref="planned_anniversary_1"/>
             <field name="state">communication</field>
             <field name="partner_field">correspondent_id</field>
-            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=1)), ("state", "=", "active"), ("child_id", "!=", False")]</field>
+            <field name="filter_domain">[("state", "=", "active"), ("child_id", "!=", False)]</field>
+            <field name="last_run" eval="datetime.now()"/>
             <field name="trigger">on_time</field>
-            <field name="trg_date_id">start_date</field>
+            <field name="trg_date_id" ref="recurring_contract.field_recurring_contract__start_date"/>
             <field name="trg_date_range">12</field>
             <field name="trg_date_range_type">month</field>
             <field name="active" eval="False"/>
@@ -52,9 +53,10 @@
             <field name="config_id" ref="planned_anniversary_3"/>
             <field name="state">communication</field>
             <field name="partner_field">correspondent_id</field>
-            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=3)), ("state", "=", "active"), ("child_id", "!=", False")]</field>
+            <field name="filter_domain">[("state", "=", "active"), ("child_id", "!=", False)]</field>
+            <field name="last_run" eval="datetime.now()"/>
             <field name="trigger">on_time</field>
-            <field name="trg_date_id">start_date</field>
+            <field name="trg_date_id" ref="recurring_contract.field_recurring_contract__start_date"/>
             <field name="trg_date_range">36</field>
             <field name="trg_date_range_type">month</field>
             <field name="active" eval="False"/>
@@ -65,9 +67,10 @@
             <field name="config_id" ref="planned_anniversary_5"/>
             <field name="state">communication</field>
             <field name="partner_field">correspondent_id</field>
-            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=5)), ("state", "=", "active"), ("child_id", "!=", False")]</field>
+            <field name="filter_domain">[("state", "=", "active"), ("child_id", "!=", False)]</field>
+            <field name="last_run" eval="datetime.now()"/>
             <field name="trigger">on_time</field>
-            <field name="trg_date_id">start_date</field>
+            <field name="trg_date_id" ref="recurring_contract.field_recurring_contract__start_date"/>
             <field name="trg_date_range">60</field>
             <field name="trg_date_range_type">month</field>
             <field name="active" eval="False"/>
@@ -78,9 +81,10 @@
             <field name="config_id" ref="planned_anniversary_10"/>
             <field name="state">communication</field>
             <field name="partner_field">correspondent_id</field>
-            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=10)), ("state", "=", "active"), ("child_id", "!=", False")]</field>
+            <field name="filter_domain">[("state", "=", "active"), ("child_id", "!=", False)]</field>
+            <field name="last_run" eval="datetime.now()"/>
             <field name="trigger">on_time</field>
-            <field name="trg_date_id">start_date</field>
+            <field name="trg_date_id" ref="recurring_contract.field_recurring_contract__start_date"/>
             <field name="trg_date_range">120</field>
             <field name="trg_date_range_type">month</field>
             <field name="active" eval="False"/>
@@ -91,9 +95,10 @@
             <field name="config_id" ref="planned_anniversary_15"/>
             <field name="state">communication</field>
             <field name="partner_field">correspondent_id</field>
-            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=15)), ("state", "=", "active"), ("child_id", "!=", False")]</field>
+            <field name="filter_domain">[("state", "=", "active"), ("child_id", "!=", False)]</field>
+            <field name="last_run" eval="datetime.now()"/>
             <field name="trigger">on_time</field>
-            <field name="trg_date_id">start_date</field>
+            <field name="trg_date_id" ref="recurring_contract.field_recurring_contract__start_date"/>
             <field name="trg_date_range">180</field>
             <field name="trg_date_range_type">month</field>
             <field name="active" eval="False"/>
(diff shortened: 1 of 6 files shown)


partner_communication_switzerland/data/sponsorship_communications_cron.xml
```diff
@@ -40,5 +31,88 @@
...
         </record>
+
+        <!-- Sponsorship anniversaries (1, 3, 5, 10 and 15 years) -->
+        <record id="sponsorship_anniversary_1_communication" model="base.automation">
+            <field name="name">Send sponsorship anniversary 1 year communication</field>
+            <field name="model_id" ref="model_recurring_contract"/>
+            <field name="config_id" ref="planned_anniversary_1"/>
+            <field name="state">communication</field>
+            <field name="partner_field">correspondent_id</field>
+            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=1))]</field>
```
ecino:
```suggestion
            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=1)), ("state", "=", "active"), ("child_id", "!=", False")]</field>
```

partner_communication_switzerland/data/sponsorship_communications_cron.xml
```diff
@@ -40,5 +31,88 @@
...
+            <field name="trg_date_range_type">month</field>
+            <field name="active" eval="False"/>
+        </record>
+        <record id="sponsorship_anniversary_3_communication" model="base.automation">
+            <field name="name">Send sponsorship anniversary 3 years communication</field>
+            <field name="model_id" ref="model_recurring_contract"/>
+            <field name="config_id" ref="planned_anniversary_3"/>
+            <field name="state">communication</field>
+            <field name="partner_field">correspondent_id</field>
+            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=3))]</field>
```
ecino:
```suggestion
            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=3)), ("state", "=", "active"), ("child_id", "!=", False")]</field>
```

partner_communication_switzerland/data/sponsorship_communications_cron.xml
```diff
@@ -40,5 +31,88 @@
...
+            <field name="trg_date_range_type">month</field>
+            <field name="active" eval="False"/>
+        </record>
+        <record id="sponsorship_anniversary_5_communication" model="base.automation">
+            <field name="name">Send sponsorship anniversary 5 years communication</field>
+            <field name="model_id" ref="model_recurring_contract"/>
+            <field name="config_id" ref="planned_anniversary_5"/>
+            <field name="state">communication</field>
+            <field name="partner_field">correspondent_id</field>
+            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=5))]</field>
```
ecino:
```suggestion
            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=5)), ("state", "=", "active"), ("child_id", "!=", False")]</field>
```

partner_communication_switzerland/data/sponsorship_communications_cron.xml
```diff
@@ -40,5 +31,88 @@
...
+            <field name="trg_date_range_type">month</field>
+            <field name="active" eval="False"/>
+        </record>
+        <record id="sponsorship_anniversary_10_communication" model="base.automation">
+            <field name="name">Send sponsorship anniversary 10 years communication</field>
+            <field name="model_id" ref="model_recurring_contract"/>
+            <field name="config_id" ref="planned_anniversary_10"/>
+            <field name="state">communication</field>
+            <field name="partner_field">correspondent_id</field>
+            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=10))]</field>
```
ecino:
```suggestion
            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=10)), ("state", "=", "active"), ("child_id", "!=", False")]</field>
```

partner_communication_switzerland/data/sponsorship_communications_cron.xml
```diff
@@ -40,5 +31,88 @@
...
+            <field name="trg_date_range_type">month</field>
+            <field name="active" eval="False"/>
+        </record>
+        <record id="sponsorship_anniversary_15_communication" model="base.automation">
+            <field name="name">Send sponsorship anniversary 15 years communication</field>
+            <field name="model_id" ref="model_recurring_contract"/>
+            <field name="config_id" ref="planned_anniversary_15"/>
+            <field name="state">communication</field>
+            <field name="partner_field">correspondent_id</field>
+            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=15))]</field>
```
ecino:
```suggestion
            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=15)), ("state", "=", "active"), ("child_id", "!=", False")]</field>
```

partner_communication_switzerland/data/sponsorship_communications_cron.xml
```diff
@@ -40,5 +31,88 @@
...
+            <field name="trigger">on_time</field>
+            <field name="trg_date_id">start_date</field>
+            <field name="trg_date_range">180</field>
+            <field name="trg_date_range_type">month</field>
+            <field name="active" eval="False"/>
+        </record>
+
+        <!-- Write and pray reminders -->
+        <t t-set="min_date" t-value="Datetime.now() - relativedelta(months=3)"/>
```
ecino: This could be done directly in the `filter_domain` field like you did for other rules

partner_communication_switzerland/data/sponsorship_communications_cron.xml
```diff
@@ -40,5 +31,88 @@
...
+        <record id="wrpr_sponsorship_reminder" model="base.automation">
+            <field name="name">Write and Pray sponsorship reminder</field>
+            <field name="model_id" ref="model_recurring_contract"/>
+            <field name="config_id" ref="sponsorship_wrpr_reminder"/>
+            <field name="state">Send Communication</field>
+            <field name="partner_field">correspondent_id</field>
+            <field name="filter_domain">[("activation_date", ">=", min_date)]</field>
```
ecino: You missed some filters that were used in the CRON. The problem is that the cron was checking if the sponsor wrote some letters with the field `sponsor_letter_ids` which is computed (therefore cannot be used in a search).

I propose that you add a field in object `recurring.contract` that will be useful for other purposes as well, that will capture the date at which the sponsor wrote his last letter. Please add the field in module `sbc_compassion` and update the date manually each time we create a new letter with direction `Supporter to Beneficiary`. 

```suggestion
            <field name="filter_domain">[("activation_date", ">=", min_date), ("state", "=", "active"), ("type", "=", "SC"), ("last_sponsor_letter", "=", False)]</field>
```

ecino (changes requested):

ecino:
I think we miss some part of your changes that you had in the other pull request that you closed (like the view and the manifest for instance). 

I also remember we talked that we can convert the waiting reminders into base action rules, please include them.

Theyiot:
> I think we miss some part of your changes that you had in the other pull request that you closed (like the view and the manifest for instance).
> 
> I also remember we talked that we can convert the waiting reminders into base action rules, please include them.

I added the missing changes.

For the second point, we ended up concluding that it was not easily feasible using automated actions. This is the reason why these modifications are not present in this _PR_.

partner_communication_switzerland/data/sponsorship_communications_cron.xml
```diff
@@ -40,5 +31,87 @@
...
         </record>
+
+        <!-- Sponsorship anniversaries (1, 3, 5, 10 and 15 years) -->
+        <record id="sponsorship_anniversary_1_communication" model="base.automation">
+            <field name="name">Send sponsorship anniversary 1 year communication</field>
+            <field name="model_id" ref="model_recurring_contract"/>
+            <field name="config_id" ref="planned_anniversary_1"/>
+            <field name="state">communication</field>
+            <field name="partner_field">correspondent_id</field>
+            <field name="filter_domain">[("start_date", ">=", Datetime.now() - relativedelta(years=1)), ("state", "=", "active"), ("child_id", "!=", False")]</field>
```
ecino: I think that I discovered the the field `last_run` is used to check if the rule was already checked, and we can use it to avoid sending the communications that were already sent in the past with CRON like this:
```suggestion
            <field name="filter_domain">[("state", "=", "active"), ("child_id", "!=", False")]</field>
            <field name="last_run" eval="Datetime.now()"/>
```
I think if we force to say that the last run was now, it will only take the new records starting from now on. Maybe you can test this and if it works change it everywhere.

partner_communication_switzerland/migrations/12.0.1.0.4/post-migration.py
```diff
@@ -0,0 +1,9 @@
+from openupgradelib import openupgrade
+
+
+def migrate(cr, installed_version):
+    if not installed_version:
+        return
+
+    openupgrade.load_xml(cr, 'partner_communication_switzerland',
+                         'data/sponsorship_communications_cron.xml')
```
ecino: This is not needed because you created new records with new xml_ids that will be loaded even if noupdate is set.

partner_communication_switzerland/models/ir_actions.py
```diff
@@ -0,0 +1,45 @@
...
+        if "records" in eval_context:
+            for partner in eval_context["records"].mapped(action.partner_field):
+                children = eval_context["records"]
+                records = self.env[model_name].search([
+                    (action.partner_field, "=", partner.id),
+                    ("id", "in", children.ids)
+                ])
+                vals = {
+                    "partner_id": partner.id,
+                    "object_ids": records.mapped("id"),
```
ecino:
```suggestion
                    "object_ids": records.ids,
```

partner_communication_switzerland/models/ir_actions.py
```diff
@@ -0,0 +1,45 @@
...
+                    "config_id": action.config_id.id,
+                    "auto_send": False,
```
ecino: We no more need to force the value as we already group the communications.
```suggestion
```

partner_communication_switzerland/models/ir_actions.py
```diff
@@ -0,0 +1,45 @@
...
+                }
+                comm_job.create(vals)
+        return False
```
ecino:
```suggestion
        return True
```
I'm more of a positive guy.

ecino (changes requested):

Theyiot:
So, I basically fixed everything you mentioned above. I tried playing around with the new parameter (`last_run`) but it is pretty tricky to test: our conditions are based on time, but the smallest time unit supported is _days_. Because of that, I was not able to really test with time conditions but I am not sure that this parameter is necessary: it looks like no communications are created retroactively. Nevertheless, to be sure that communications are not created after a certain amount of time, I still left these parameters in the code.
The rest of the modifications are straightforward, so they won't be covered with more details.

ecino (approved):

PR #1218, opened 2020-08-17
CO-3385 add security groupy for accessing criminal records

ecino:
CO-3385 refactor compute of criminal records and security group

CO-3385 swap get_file_type method placement

CO-3385 refactored inverse of criminal record for access of event registrations

CO-3385 changed notification behaviour and added option in settings

Replaces #1189

### DIFF
--- a/partner_compassion/models/partner_compassion.py
+++ b/partner_compassion/models/partner_compassion.py
@@ -10,6 +10,7 @@
 import logging
 import tempfile
 import uuid
+import base64
 
 from odoo import api, registry, fields, models, _
 from odoo.tools import mod10r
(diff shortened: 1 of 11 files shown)


partner_compassion/models/partner_compassion.py
```diff
@@ -280,6 +315,8 @@ def write(self, vals):
         email = vals.get("email")
         if email:
             vals["email"] = email.strip()
+        if vals.get("criminal_record"):
+            vals["criminal_record_date"] = fields.Date.today()
```
Theyiot, line 319: The field `criminal_record_date` will get override every time the write method is called here (if a partner has a criminal record set). Is this really the behaviour that is expected ?

website_event_compassion/__manifest__.py
```diff
@@ -29,7 +29,7 @@
 # pylint: disable=C8101
 {
     "name": "Compassion Events Website",
-    "version": "12.0.1.0.2",
+    "version": "12.0.1.0.3",
```
Theyiot, line 32: The version is new but there is no _pre_ or _post-migration_ script, is that normal ?

Theyiot (changes requested):
The changes globally look good to me. There are just two points with which I am not quite sure. If you can take a look @ecino, that would be great !

partner_compassion/models/partner_compassion.py
ecino: `criminal_record` will be present in `vals` only if it is updated so it won't change each time we write other fields in the partner. This is what we want in order to consider that each time we update it, it means we put a more recent document.

website_event_compassion/__manifest__.py
ecino: Initially I wanted to put the migration here, but eventually I put it in `partner_compassion`. However it's not bad to increment the module version when we make changes.

PR #1219, opened 2020-08-20
CO-3431 Amount on payment slip and scan_line is now the right amount

Theyiot:


ecino (changes requested):
You should add a migration script to force reload `report_compassion/report/bvr_sponsorship.xml`

ecino:
I think you forgot this PR.

Theyiot:
> I think you forgot this PR.

You're right ;) The script has now been added

ecino (approved):

PR #1224, opened 2020-08-31
CO-3437 Update user email after partner update

nfluckiger:


ecino:
With this change we risk to update the user's email without changing the one in the partner (if we don't save). I think we need to discuss the task, I'm not sure about it.

PR #1227, opened 2020-09-08
CO-3260 Number of magazines is now a selection field instead of an int

Theyiot:


partner_compassion/models/partner_compassion.py
```diff
@@ -87,7 +87,25 @@ class ResPartner(models.Model):
...
+            ("six", "6"),
+            ("seven", "7"),
+            ("eight", "8"),
+            ("nine", "9"),
+            ("ten", "10"),
+            ("fifteen", "15"),
+            ("twenty", "20"),
+            ("twenty_five", "25"),
+            ("fifty", "50"),
+        ], string="Number of Magazines", required=True)
```
ecino:
```suggestion
        ], string="Number of Magazines", required=True, default="one")
```

ecino (changes requested):

PR #1229, opened 2020-09-10
CO-3284 Pending state is now replaced by an activity

Theyiot:
We simply remove legacy inheritance according to other modifications from here: https://github.com/CompassionCH/compassion-modules/pull/1447

### DIFF
--- a/sponsorship_switzerland/models/contracts.py
+++ b/sponsorship_switzerland/models/contracts.py
@@ -193,9 +193,7 @@ class RecurringContracts(models.Model):
         Update partner to add the 'Sponsor' category
         """
         super().contract_active()
-        # TODO CO-3284 Replace this check by looking if mail.activity exists
-        need_validation = self.filtered(lambda s: s.partner_id.state != "active")
-        if need_validation:
+        if self.partner_id.activity_ids or self.correspondent_id.activity_ids:
             raise UserError(
                 _("Please verify the partner before validating the sponsorship")
             )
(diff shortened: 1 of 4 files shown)


ecino (changes requested):
Please remove the `state` field inside `partner_compassion.py` and all related usages.

ecino (changes requested):
You should still replace remove the fields from the view `partner_compassion_view.xml`

Theyiot:
> You should still replace remove the fields from the view `partner_compassion_view.xml`

Last commit was amended with all TODOs removed.

sponsorship_switzerland/models/contracts.py
```diff
@@ -193,9 +193,7 @@ def contract_active(self):
         Update partner to add the 'Sponsor' category
         """
         super().contract_active()
-        # TODO CO-3284 Replace this check by looking if mail.activity exists
-        need_validation = self.filtered(lambda s: s.partner_id.state != "active")
-        if need_validation:
+        if self.partner_id.activity_ids or self.correspondent_id.activity_ids:
```
ecino: The method may be multi (self is a recordset) therefore it's forbidden to use `self.partner_id` or `self.correspondent_id`.
You can use the `mapped` method to circumvent this problem.
```suggestion
        if self.mapped("partner_id.activity_ids") or self.mapped("correspondent_id.activity_ids"):
```

ecino (changes requested):

sponsorship_switzerland/models/contracts.py
Theyiot: I (wrongly) assumed these methods would be called on singleton recordset. This is fixed by next commit.

ecino (approved):

PR #1232, opened 2020-09-18
CO-3256 Update sms sponsorship to be mobile sponsorship

nfluckiger:


sms_939/models/sms_child_request.py
```diff
@@ -68,7 +68,8 @@ def _take_child_from_website(self):
             self.write({"child_id": child.id, "state": "child_reserved"})
             child.hold_id.sms_request_id = self.id
             child.remove_from_wordpress()
-            child.write({"state": "S"})
+            if self.type == "sms":
```
ecino, line 71: This would be needed for all request types. You could as well change the state name in the child from "Reserved for SMS" to "Reserved for Mobile".

ecino (changes requested):

sms_939/data/sms_cron.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
+        <record id="mobile_sponsorship_revoke" model="base.automation">
+            <field name="name">Mobile app Sponsorship revoker</field>
+            <field name="model_id" ref="model_sms_child_request"/>
+            <field name="trigger">on_time</field>
+            <field name="trg_date_id" ref="field_sms_child_request_date"/>
+            <field name="filter_domain">[('type', '=', 'mobile')]</field>
```
ecino: We only want to cancel pending requests.
```suggestion
            <field name="filter_domain">[('type', '=', 'mobile'),('state','in',['new','child_reserved'])]</field>
```

ecino (changes requested):

ecino (approved):

PR #1242, opened 2020-11-02
Crowdfunding fixes

ecino:


crowdfunding_compassion/forms/project_update_form.py
```diff
@@ -74,4 +78,27 @@ def form_before_create_or_update(self, values, extra_values):
...
+            if notify_ids:
+                user = self.env["res.partner"].sudo().browse(notify_ids[0][2]) \
+                    .mapped("user_ids")[:1]
+                self.main_object.sudo().activity_schedule(
+                    'mail.mail_activity_data_todo',
+                    summary="Verify project information.",
+                    note=f"{self.main_object.project_owner_id.name} updated the "
+                    f"name and description of the project. "
+                    f"Please check if the information is good enough.",
+                    user_id=user.id
```
Theyiot, line 100: Are we sure that there will be a user in all cases (since we get the user using `:1` instead of simply `0`) ? I don't know if it would be overkill to put the `activity_schedule()` inside a `if user:`.

Theyiot (approved):
Really good modifications and fixes. You did catch quite a few subtle issues !

I simply added a comment on a little concern I have, otherwise it looks good to me. I did not check the translations, though.

crowdfunding_compassion/forms/project_update_form.py
ecino: If notify_ids is set there should be no reason that there is no user so I think we can keep it like that. Thanks for checking.

PR #1244, opened 2020-11-04
CO-3482 Adding child's information in My Children

Theyiot:
In order for it to work properly, one needs to add an image named `guy.png` and one called `lady.png` inside the `website_compassion/static/src/img/` folder.

Other PR are waiting on this one (since they are based on it), so it has the highest importance.


ecino:
Please tell me when it's ready. Otherwise you can as well create draft pull requests and then make it active when they are ready.

Theyiot:
> Please tell me when it's ready. Otherwise you can as well create draft pull requests and then make it active when they are ready.

I was waiting for a call to determine whether some modifications were required or not. It seems to be complete, the fields may change a bit (naming, what is shown or not), but the PR is now ready. These modifications can be done afterward easily and they won't impact the overall layout.

ecino:
"In order for it to work properly, one needs to add an image named guy.png and one called lady.png inside the website_compassion/static/src/img/ folder."

=> Can't you add them to your pull request? We have them in crowdfunding_compassion module already if they are the same.

ecino (approved):

PR #1246, opened 2020-11-06
CO-3485 Adding children's pictures in the view, with download link

Theyiot:
The pictures of the children are now displayed inside the view and it is possible to download them in three different ways:
- only one single image of a child
- all the images of a given child
- all the images of all our children
If there are more than one image, then a _.zip_ archive is created with all the images inside.

Most of the modifications are already done in PR #1244, which is ready.

### DIFF
--- a/website_compassion/controllers/my_account.py
+++ b/website_compassion/controllers/my_account.py
@@ -6,8 +6,118 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
+from os import path, remove
+from zipfile import ZipFile
+from urllib.request import urlretrieve, urlopen
+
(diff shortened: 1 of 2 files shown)


ecino:
Apparently the same commits appear, maybe you did a rebase ? You should now rebase this branch onto my_account branch in order to keep only the new commit.

Theyiot:
> Apparently the same commits appear, maybe you did a rebase ? You should now rebase this branch onto my_account branch in order to keep only the new commit.

I created the PR before the other branch was merged. The rebase has been applied now, the PR is now ready.

ecino (approved):

PR #1247, opened 2020-11-06
CO-3483 Adding children's letters history

Theyiot:
The complete history of the letters has been added in the view My Children. There is the possibility to view or to download the letters individually.

It could be nice to add the possibility to add the possibility to download all letters (from one child and maybe all letters) using a simple click (as for pictures). But the good functioning of the _pdf_ preview and download must be first tested on a complete database.


website_compassion/controllers/my_account.py
```diff
@@ -153,6 +152,14 @@ def my_children(self, **kwargs):
...
         partner = request.env.user.partner_id
         children = _get_user_children()
+        child = children.filtered(lambda c: c.id == child_id)
+        sponsorship_ids = child.sponsorship_ids.filtered(
+            lambda contract: contract.partner_id == partner or
+            contract.correspondent_id == partner
+        )
+        letter_ids = request.env["correspondence"].search([
+            ("sponsorship_id", "=", sponsorship_ids[0].id),
+        ])
```
ecino: No need to go through children to find the letters.
```suggestion
       letters = request.env["correspondence"].search([
            ("partner_id", "=", partner.id),
            ("child_id", "=", child_id)
        ])
```
I would suggest not to name 'letter_ids' as it holds a recordset and not a list of ids.

ecino (changes requested):

website_compassion/controllers/my_account.py
Theyiot: You're right, I modified the name and will correct in all templates the `_ids` of recordset in the future,

PR #1249, opened 2020-11-09
CO-3484 Gift history integration

Theyiot:
The gift history is now displayed in the view _My Children_. A link to the donations of the website as also been added.

The `id` of the category (Sponsorship gifts) has been hard coded, because the name is translated and hence was cumbersome to use.

For this to work, the portal group must be granted access to the model `product.product` as well as `product.category`.

### DIFF
--- a/website_compassion/__manifest__.py
+++ b/website_compassion/__manifest__.py
@@ -35,6 +35,7 @@
     "license": "AGPL-3",
     "website": "http://www.compassion.ch",
     "data": [
+        "security/ir.model.access.csv",
         "template/my_account_components.xml",
         "template/my_account_personal_info.xml",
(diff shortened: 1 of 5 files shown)


ecino:
"For this to work, the portal group must be granted access to the model `product.product` as well as `product.category`."

=> Would be good if you provide the security accesses within the security folder, so that it works out of the box.
=> Conflicts were introduced with other branches

Theyiot:
> "For this to work, the portal group must be granted access to the model `product.product` as well as `product.category`."
> 
> => Would be good if you provide the security accesses within the security folder, so that it works out of the box.
> => Conflicts were introduced with other branches

The changes have been added !

ecino (approved):

PR #1255, opened 2020-11-18
CO-3475 fix forget me

jonathantarabbia:
- execute actions with grant permission sudo()
- clear birthdate, tags, global_id, reference

### DIFF
--- a/partner_compassion/models/partner_compassion.py
+++ b/partner_compassion/models/partner_compassion.py
@@ -529,9 +529,10 @@ class ResPartner(models.Model):
                 "partner_latitude": False,
                 "partner_longitude": False,
                 "birthdate_date": False,
+                "category_id": [(5, 0, 0)]
             }
         )
-        self._cr.execute("""update res_partner set ref='', global_id='' where id=""" + str(self.id))
+        self._cr.execute("update res_partner set ref=NULL, global_id=NULL where id=%s", [self.id])
         self.advocate_details_id.sudo().unlink()
         self.survey_inputs.sudo().unlink()
         self.env["mail.tracking.email"].sudo().search([("partner_id", "=", self.id)]).unlink()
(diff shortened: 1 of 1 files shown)


partner_compassion/models/partner_compassion.py
```diff
@@ -528,18 +528,21 @@ def forget_me(self):
...
                 "partner_latitude": False,
                 "partner_longitude": False,
+                "birthdate_date": False,
             }
         )
-        self.advocate_details_id.unlink()
-        self.survey_inputs.unlink()
-        self.env["mail.tracking.email"].search([("partner_id", "=", self.id)]).unlink()
-        self.env["auditlog.log"].search(
+        self._cr.execute("""update res_partner set ref='', global_id='' where id=""" + str(self.id))
```
ecino: This difference avoids SQL injection
```suggestion
        self._cr.execute("update res_partner set ref=NULL, global_id=NULL where id=%s", [self.id])
```

partner_compassion/models/partner_compassion.py
```diff
@@ -528,18 +528,21 @@ def forget_me(self):
...
+        self.env["auditlog.log"].sudo().search(
             [("model_id.model", "=", "res.partner"), ("res_id", "=", self.id)]
         ).unlink()
-        self.env["partner.communication.job"].search(
+        self.env["partner.communication.job"].sudo().search(
             [("partner_id", "=", self.id)]
         ).unlink()
-        self.message_ids.unlink()
+        self.message_ids.sudo().unlink()
+        self.category_id.sudo().unlink()
```
ecino: This is wrong, you will delete the tag instead of deleting the relation! I propose to add this above instead.
```suggestion
```

partner_compassion/models/partner_compassion.py
```diff
@@ -528,18 +528,21 @@ def forget_me(self):
                 "geo_point": False,
                 "partner_latitude": False,
                 "partner_longitude": False,
+                "birthdate_date": False,
```
ecino, line 531:
```suggestion
                "birthdate_date": False,
                "category_id": [(5, 0, 0)]
```

ecino (changes requested):

ecino (approved):

PR #1260, opened 2020-11-25
CO-3533 - fix notification donations

jonathantarabbia:
- now, the notification system send product names when the partner made several donations.


ecino:
Doesn't it produce too long texts for the notification or is it not a problem?

ecino (approved):

PR #1262, opened 2020-11-26
CO-3519 - Improvements website

jonathantarabbia:
- now the sitemap.xml file is managed with a new no_sitemap attribute at the HTTP routes level to exclude the route in the sitemap.xml file
- example of how to create paths in sitemap.xml file for projects
- project management by website with access management to a project from another website
- improvements
- add filter and pager on list page project


crowdfunding_compassion/controllers/projects_controller.py
```diff
@@ -7,31 +7,77 @@
...
+                total = project_obj.search_count(filters)
+
+                pager = request.website.pager(
+                    url='/projects',
+                    total=total,
+                    page=page,
+                    step=self._project_post_per_page,
+                    url_args={'type': type},
+                )
+                # TODO connect pagination to backend -> CO-3213
```
ecino: I guess this is done and can now be removed (and we can close the issue).

crowdfunding_compassion/controllers/projects_controller.py
```diff
@@ -7,31 +7,77 @@
...
+                                             limit=self._project_post_per_page, page=page)
+                return request.render(
+                    "crowdfunding_compassion.project_list_page",
+                    {"projects": projects,
+
+                     "type": type,
+                     "pager": pager},
+                )
+            else:
+                raise werkzeug.exceptions.BadRequest()
```
ecino: This way we will raise the error for any other situation where it didn't return.
```suggestion
        raise werkzeug.exceptions.BadRequest()
```

crowdfunding_compassion/controllers/projects_controller.py
```diff
@@ -7,31 +7,77 @@
...
+        for f in projects.search(dom):
+            loc = '/project/%s' % slug(f)
+            if not qs or qs.lower() in loc:
+                yield {'loc': loc}
+
+    @route(["/projects", "/projects/page/<int:page>"], auth="public", website=True, sitemap=sitemap_projects)
+    def get_projects_list(self, type=None, page=1, year=None, **opt):
+        if request.website:
+            website_id = request.website.id
+            if website_id != 1:  # only for other site
```
ecino: This way we match precisely the website
```suggestion
            if website_id == request.env.ref(
                    "crowdfunding_compassion.crowdfunding_website").sudo().id
```

crowdfunding_compassion/controllers/projects_controller.py
```diff
@@ -96,7 +142,8 @@ def project_creation(self, page=1, project_id=0, **kwargs):
            auth="public",
            type="http",
            method='POST',
-           website=True, noindex=['robots', 'meta', 'header'])
+           website=True, noindex=['robots', 'meta', 'header'],
```
ecino: I think we can remove the noindex parameters and we will deinstall the problematic module.
```suggestion
           website=True,
```

crowdfunding_compassion/models/crowdfunding_project.py
```diff
@@ -349,33 +353,35 @@ def toggle_website_published(self):
...
+        # finished_projects = self.env[self._name]
+        # if not limit or (limit and len(active_projects) < limit):
+        #     finish_limit = limit - len(active_projects) if limit else None
+        #     finished_projects = self.search(
+        #         [
+        #             ("deadline", "<", date.today()),
+        #         ] + filters, offset=offset, limit=finish_limit, order="deadline DESC"
+        #     )
+
+        return active_projects # + finished_projects
```
ecino: We should have a page where we can see as well the finished projects, is this no longer possible ? It would be good that you add this as a quick filter on the "projects page" in order to display the finished projects.

crowdfunding_compassion/models/website.py
```diff
@@ -0,0 +1,16 @@
...
+    def rule_is_enumerable(self, rule):
+        """ Create a route parameter named no_sitemap. Set it
+            to true to exclude this route from the sitemap
+            generation.
+            :type rule: werkzeug.routing.Rule
+            :rtype: bool
+        """
+        if rule.endpoint.routing.get('no_sitemap'):
+            return False
+        return super(Website, self).rule_is_enumerable(rule)
```
ecino: Can you move this file inside module `compassion-modules/cms_form_compassion`? Like that we will have the attribute available for all our website modules and not only the crowdfunding.

ecino (changes requested):

ecino (approved):

PR #1269, opened 2020-12-01
CO-3481 Adding my information page

Theyiot:
Here are added the functionalities for the page _My information_ to work. Even though a lot of the functionalities were almost completely implemented in other modules, it took really long to take these changes, make them work (or redo them when they did not meet expectations) and ensure proper functioning.

There is still one problem: the first form does not close on save, even though no error is logged and the modifications are actually applied (if you refresh the page). It seems to me that it is implemented the same way than the other that work properly (_communication preferences_), so I find myself out of ideas... I would be really grateful if @ecino could take a look, at least to give me an idea of how to solve this (I searched in the routes but everything looks fine).

This is parallel to: https://github.com/CompassionCH/compassion-modules/pull/1483


partner_compassion/models/partner_compassion.py
```diff
@@ -318,16 +318,6 @@ def create(self, vals):
...
-        email = vals.get("email")
-        if email:
-            vals["email"] = email.strip()
-            # Push email to non-internal users
-            user_ids = self.mapped("user_ids").filtered("share").ids
-            self.env.cr.execute("""
-                UPDATE res_users
-                SET login=%s
-                WHERE id=ANY(%s)
-            """, [vals["email"], user_ids])
```
ecino: Do you replace this with a special form where they can change their login ?

website_compassion/controllers/my_account.py
```diff
@@ -196,8 +198,54 @@ def my_child(self, child_id=None, **kwargs):
...
+        values = self._prepare_portal_layout_values()
+        values.update({
+            "partner": partner,
+            "coordinates_form": coordinates_form,
+            "delivery_form": delivery_form,
+        })
+
+        # This fixes an issue that forms fail after first submission
+        if form_success:
+            result = request.redirect("/my/information")
```
ecino, line 228: Maybe this is no more needed, did you try without? I guess we should be able to work with forms without this if we understood well how they work.
Maybe you also removed some javascript that would be useful to handle various forms in the same page (in popups).

ecino:
Now that I merged the other branch there are some conflicts.

partner_compassion/models/partner_compassion.py
Theyiot: No, there is a `write` of the `login` already in _MailChimp_ (that was not there previously). Because of that duplication, modifying the login would fail because of concurrent update in the database.

ecino: There is no such thing in mailchimp module, or can you point me where it is?

website_compassion/controllers/my_account.py
Theyiot: No, this does not work properly if I remove this part of the code (a manual refresh of the page is required to see the changes).

partner_compassion/models/partner_compassion.py
Theyiot: You were right, I added the code back and fixed the errors so that it works properly again.

eicher31 (approved):
At first glance, it doesn't seem bad to me! I will test on the devel.

eicher31:
@Theyiot is it possible to fix pylint and flake8 errors?

PR #1282, opened 2021-02-23
CO-3578 create model for partner segmentation and add/adapt views

JonathanGuerne:
Add 2 new models 

- res.partner.segment
- res.partner.segment.affinity

A segment categorized partner (link them with different habits). Segment_Affinity is used to link a partner with a Segment over a certain affinity (percentage)


partner_compassion/models/parter_segment_affinity.py
```diff
@@ -0,0 +1,135 @@
...
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+from odoo import api, fields, models
+from math import exp
+
+
+class ResPartnerSegmentAffinity(models.Model):
+    _name = "res.partner.segment.affinity"
+
```
ecino, line 18: - Add a description attribute in new models to give a human readable name of the model.
- Add default order attribute to ensure records are read in a given order
```suggestion
_description = "Partner segment affinity"
_order = "partner_id desc, affinity desc"

```

partner_compassion/models/parter_segment_affinity.py
```diff
@@ -0,0 +1,135 @@
...
+    partner_id = fields.Many2one("res.partner", "Partner")
+    segment_id = fields.Many2one("res.partner.segment", "Segment")
+
+    affinity = fields.Float(
+        help='affinity of the partner for this segment (in percentage).')
```
ecino: We conventionnaly use double quotes for all strings in the code.
```suggestion
        help="affinity of the partner for this segment (in percentage).")
```

partner_compassion/models/parter_segment_affinity.py
```diff
@@ -0,0 +1,135 @@
...
+    ]
+
+    @api.model
+    def create(self, vals):
+        """
+        Create a segment_affinity instance. Update partner reference on segment affinity
+        :param vals: values used for model creation
+        :return: id of the newly created model
+        """
+        seg_affin_id = super(ResPartnerSegmentAffinity, self.sudo()).create(vals)
```
ecino: Naming conventions on our variables : I would remove the `_id` suffix, because the `create` method returns directly the recorded, not only the `id` of it.
```suggestion
        seg_affin = super(ResPartnerSegmentAffinity, self.sudo()).create(vals)
```

partner_compassion/models/parter_segment_affinity.py
```diff
@@ -0,0 +1,135 @@
...
+
+        # update partner.all_segment_affinity will trigger primary and secondary segment computation
+        # seg_affin_id.partner_id.write({'all_segments_affinity': [(4, seg_affin_id.id)]})
+        seg_affin_id.partner_id.all_segments_affinity = [(4, seg_affin_id.id)]
+        # seg_affin_id.partner_id._compute_prim_sec_segments()
+        return seg_affin_id
```
ecino: Please remove old code
```suggestion
        seg_affin.partner_id.all_segments_affinity = [(4, seg_affin.id)]
        return seg_affin
```

partner_compassion/models/parter_segment_affinity.py
```diff
@@ -0,0 +1,135 @@
...
+    @api.multi
+    def unlink(self):
+        """
+        Unlink insance of the model. Update partner reference.
+        :return: True
+        """
+        for seg_affin in self:
+            seg_affin.partner_id.all_segments_affinity = [(2, seg_affin.id)]
+            super(ResPartnerSegmentAffinity, seg_affin).unlink()
+        return True
```
ecino: Just a short variant to unlink all at once.
```suggestion
        return super().unlink()
```

partner_compassion/models/parter_segment_affinity.py
```diff
@@ -0,0 +1,135 @@
...
+        :param partner_id: the partner related to the answers
+        :return: True (update partner all_segments_affinity)
+        """
+
+        # simplified explanation of the segmentation percentage computation
+        # answer_array * matrix -> scaled matrix ( + constant) -> segmentation score
+        # -> segmentation percentage (= segmentation affinity)
+
+        vals = {
+            'partner_id': partner_id.id
```
ecino: You only use `partner_id.id` so I would suggest you directly pass this as parameter when you call the method and here you just use it like this.
```suggestion
            'partner_id': partner_id
```

partner_compassion/models/survey.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def write(self, vals):
+        out = super(SurveyUserInput, self).write(vals)
+
+        # compute segment affinity if needed
+        state = 'state' in vals and vals['state'] == 'done'
+        _id = self.survey_id == self.env.ref('partner_compassion.partner_segmentation_survey')
+        if state and _id:
+            ans = self._get_answer_as_array()
+            self.env['res.partner.segment.affinity'].segment_affinity_engine(ans, self.partner_id)
```
ecino: You can directly pass the id as you don't need the whole record.
```suggestion
            self.env['res.partner.segment.affinity'].segment_affinity_engine(ans, self.partner_id.id)
```

partner_compassion/models/survey.py
```diff
@@ -0,0 +1,114 @@
...
+    order_in_question = fields.Integer(
+        'Order of the question. Used in multiple options question when order is important. should starts at 0',
+    )
+
+
+class SurveyUserInput(models.Model):
+    _inherit = "survey.user_input"
+
+    def write(self, vals):
+        out = super(SurveyUserInput, self).write(vals)
```
ecino: super parameters are optional when you are calling upon the same self instance.
```suggestion
        res = super().write(vals)
```

partner_compassion/models/survey.py
ecino: - Beware that the `write` method can be called on a recordset. You must iterate on it to handle all records properly.
- Use variable names that are human understandable
- Use double quotes for strings
- Don't use variables if the value is used only once, except if it makes it more understandable what it is.
```suggestion
        is_done = vals.get("state") == "done"
        segment_survey = self.env.ref("partner_compassion.partner_segmentation_survey")
        for user_input in self:
            if user_input.survey_id == segment_survey and is_done:
                ans = user_input._get_answer_as_array()
                self.env["res.partner.segment.affinity"].segment_affinity_engine(ans, user_input.partner_id.id)
```

partner_compassion/models/partner_segment.py
```diff
@@ -0,0 +1,53 @@
...
+##############################################################################
+from odoo import fields, models
+
+
+class ResPartnerSegment(models.Model):
+    _name = "res.partner.segment"
+    _description = "Partner Segmentation"
+
+    segment_index = fields.Integer(string='Segment Index used for segment computation by segmentation engine')
+    name = fields.Char(string='Name')
```
ecino:
```suggestion
    name = fields.Char(required=True)
```

partner_compassion/models/partner_segment.py
```diff
@@ -0,0 +1,53 @@
...
+    main_driver = fields.Char(string='Main driver', help='short description of the segment')
```
ecino:
```suggestion
    main_driver = fields.Char()
```

partner_compassion/models/partner_segment.py
```diff
@@ -0,0 +1,53 @@
...
+    segment_total = fields.Integer(
+        help='percentage of categorized partners for whom this is the primary segment',
+        compute='_compute_segment_total'
+    )
+
+    survey_result = fields.Html()
+    image = fields.Binary(help='segment illustration')
+
+    primary_partners = fields.One2many('res.partner', compute='_compute_primary_partners')
+    secondary_partners = fields.One2many('res.partner', compute='_compute_secondary_partners')
```
ecino: Naming conventions : relational fields should be suffixed to indicate it.
```suggestion
    primary_partner_ids = fields.One2many(
        "res.partner", "Primary partners", compute="_compute_primary_partner"')
    secondary_partner_ids = fields.One2many(
        "res.partner", "Secondary partners", compute="_compute_secondary_partners")
```

partner_compassion/models/partner_segment.py
```diff
@@ -0,0 +1,53 @@
...
+
```
ecino, line 41: We can add `sql_constraints` to add some data validation:
```suggestion

_sql_constraints = [("unique_index", "unique(segment_index)", "Segment index must be unique")]
```

partner_compassion/models/partner_compassion.py
```diff
@@ -281,6 +299,24 @@ def _compute_thankyou_letter(self):
...
+                partner.primary_segment_id = affinity_for_segments[0][1].segment_id
+                partner.secondary_segment_id = affinity_for_segments[1][1].segment_id
+
+            elif affinity_for_segments and len(affinity_for_segments) > 0:
+                partner.primary_segment_id = affinity_for_segments[0][1].segment_id
+                partner.secondary_segment_id = None
+
+            else:
+                partner.primary_segment_id = None
+                partner.secondary_segment_id = None
```
ecino: - You can rely on the default order of recordsets to achieve the same result (see my comment in the segment affinity class).
- You can rely on python slicing to get given elements in the recordset even if the recordset is empty. It will return an empty recordset and the `.segment_id` will return `False`
```suggestion
            partner.primary_segment_id = partner.all_segments_affinity[:1].segment_id
            partner.secondary_segment_id = partner.all_segments_affinity[1:2].segment_id
```

partner_compassion/models/partner_segment.py
```diff
@@ -0,0 +1,53 @@
...
+#
+##############################################################################
+from odoo import fields, models
+
+
+class ResPartnerSegment(models.Model):
+    _name = "res.partner.segment"
+    _description = "Partner Segmentation"
+
+    segment_index = fields.Integer(string='Segment Index used for segment computation by segmentation engine')
```
ecino: - String is usually the label in the view so you must keep it short. Use the help parameter to give more insight.
- String is the first attribute of field and can be omitted : in that case the label will be as the field_name => Field Name
- Add some other parameters
- Lines should be 89 characters long at most
```suggestion
    segment_index = fields.Integer(
        required=True,
        help="Used for segment computation by segmentation engine"
    )
```

ecino (changes requested):
I didn't comment everywhere but you can take the principles I listed and correct everywhere that is relevant.

Functionality looks good, only cosmetic/conventions issues to be adressed.

Thanks for the contribution :+1:

ecino (approved):

PR #1284, opened 2021-03-02
CO-3559 Prevent Sponsorship if check duplicate activity is not done.

coup1234:
- new activity type called Duplicate Check activity.
- A Duplicate Check activity is created when a partner is created from
  cms_form_compassion (to be tested if possible).
- Sponsorship is not validate only if a check duplicate activity is active.


sponsorship_switzerland/models/contracts.py
```diff
@@ -268,11 +270,19 @@ def _check_sponsorship_is_valid(self):
...
+                or self.mapped("correspondent_id.activity_ids")\
+                .filtered(lambda l: l.activity_type_id.id == check_duplicate_activity_id):
             raise UserError(
                 _("Please verify the partner before validating the sponsorship")
             )
+        # if self.mapped("partner_id.activity_ids") or self.mapped(
+        #         "correspondent_id.activity_ids"):
+        #     raise UserError(
+        #         _("Please verify the partner before validating the sponsorship")
+        #     )
```
ecino: I prefer that we remove old code instead of commenting it. Git already enables us to go back in history if needed and thus we keep our clode clean.

ecino (changes requested):
I think you could do the change directly in `compassion-modules/cms_form_compassion` module in order to avoid overriding the `match_create` method here. You will just need to make one more pull request on the compassion-modules repository.

ecino (approved):

PR #1286, opened 2021-03-03
CO-3560 split mailchimp update jobs. avoid duplicates or call on non …

JonathanGuerne:
…changed partners

left old job untouched as it seems to be call periodically on all partners.

(sorry for pr on wrong branch)


mass_mailing_switzerland/models/mailchimp_lists.py
```diff
@@ -7,14 +7,86 @@
...
         # Avoid loops in mailchimp sync
-        return super(MailchimpLists,
-                     self.with_context(import_from_mailchimp=True)
-                     ).process_member_from_stored_response(pending_record)
+        out = super(MailchimpLists,
+                    self.with_context(import_from_mailchimp=True)
+                    ).process_member_from_stored_response(pending_record)
+
+        # mailing_contact with invalid_email should be unlinked from res.partner
+        invalid_contacts = self.env['mail.mass_mailing.contact'].search([('is_email_valid', '=', False)])
```
ecino: I don't get the logic behind this. The method `process_member_from_stored_response` is called when we fetch a member info from Mailchimp. Then why checking all contacts at this time for invalid email? I think either here we should only process the members that were fetched (instead of all mailing contacts), or we should move this code at the place where `is_email_valid` value is changed.

mass_mailing_switzerland/models/mailchimp_lists.py
```diff
@@ -7,14 +7,86 @@
...
+
+        for invalid_contact in invalid_contacts:
+
+            ref_partner = self.env['res.partner'].search([('id', '=', invalid_contact.partner_id.id)])
```
ecino:
```suggestion
            ref_partner = invalid_contact.partner_id
```

mass_mailing_switzerland/models/mailchimp_lists.py
```diff
@@ -7,14 +7,86 @@
...
+                        {
+                            "config_id": invalid_comm.id,
+                            "partner_id": ref_partner.id,
+                            "object_ids": ref_partner.id,
+                        }
+                    )
+                ref_partner.message_post(
+                    body=_("The email couldn't be sent due to invalid address"),
+                    subject=ref_partner.invalid_mail
+                )
```
ecino: We didn't try to send an email at this point, so this message is inappropriate. I would change to "Mailchimp detected an invalid email address").

mass_mailing_switzerland/models/mailchimp_lists.py
```diff
@@ -7,14 +7,86 @@
...
+            else:
+                # if reference on partner is not there but mailing contact still exist remove mailing contact
+                invalid_contact.unlink()
+        return out
+
+    @api.multi
+    def members_cleaning_cron(self):
+        """
+        Cron used to clean mailchimp list of members. Mailchimp might store contact information
```
ecino: Is it a problem to have contacts in Mailchimp that are not in Odoo? What problem are you solving here? I'm a bit afraid to archive like this a potential high numbers of people from Mailchimp. When we archive them we cannot undo the operation easily in case of a mistake. This whole procedure looks risky to me...

mass_mailing_switzerland/models/mass_mailing_contact.py
```diff
@@ -42,6 +42,34 @@ def get_partner(self, email):
...
+    @api.multi
+    def write(self, values):
+        out = super().write(values)
+        # can't be simplified because we are looking for is_email_valid == False
+        # AND is_email_valid is in values (return None otherwise)
+        if values.get('is_email_valid') is False:
+            for mailing_contact in self:
+                # assure is_valid_email as not been reset (compute field might have been trigger)
+                if mailing_contact.is_email_valid:
+                    mailing_contact.is_email_valid = False
```
ecino: I think the underlying problem is that `is_email_valid` is a compute field and the mailchimp module writes into that field but it's kind of wrong. To fix this better, we should simply convert the field to a regular field, by redefining the field and set the attribute `compute=False`.

ecino: This change should be directly made into `paid-addons/mailchimp` module (because it's already wrong at this stage)

mass_mailing_switzerland/models/mass_mailing_contact.py
```diff
@@ -79,8 +107,30 @@ def action_update_to_mailchimp(self):
         """
         if self.env.context.get("skip_mailchimp"):
             return True
-        return super(MassMailingContact,
-                     self.with_context(lang="en_US")).action_update_to_mailchimp()
+
+        for contact_to_update in self:
+            # if previous write failed reference to mailchimp member will be lost. error 404
+            try:
+                return super(MassMailingContact,
```
ecino: You shouldn't return here, otherwise you will exit the method at the first iteration.
```suggestion
                super(MassMailingContact,
```

mass_mailing_switzerland/models/mass_mailing_contact.py
```diff
@@ -79,8 +107,30 @@ def action_update_to_mailchimp(self):
...
+                             contact_to_update.with_context(lang="en_US")).action_update_to_mailchimp()
+            except Exception as e:
+
```
ecino: When exceptions are raised, we could have closed cursors to database and transactions are rolled back, leading to potential new errors if we still use them. We should clear the environment to restore connection to database.
```suggestion
            self.env.clear()
```

mass_mailing_switzerland/models/mass_mailing_contact.py
```diff
@@ -79,8 +107,30 @@ def action_update_to_mailchimp(self):
...
+                available_mailchimp_lists = self.env['mailchimp.lists'].search([])
+                lists = available_mailchimp_lists.mapped('odoo_list_id').ids
```
ecino: This should be moved inside the `if` to avoid unnecessary computations when we are dealing with another exception.

mass_mailing_switzerland/models/mass_mailing_contact.py
```diff
@@ -79,8 +107,30 @@ def action_update_to_mailchimp(self):
...
+                # solution : we remove previous link to mailchimp and export the contact with new mail
+                if e.args[0] and literal_eval(e.args[0])['status'] == 404:
+                    contact_to_update.subscription_list_ids.filtered(
+                        lambda x: x.list_id.id in lists).write({"mailchimp_id": False})
+                # raise exception if it's any other type
+                else:
+                    raise e
+
+                # once link is remove member can again be exported to mailchimp
+                return super(MassMailingContact,
```
ecino:
```suggestion
                super(MassMailingContact,
```

mass_mailing_switzerland/models/mass_mailing_contact.py
```diff
@@ -42,6 +42,34 @@ def get_partner(self, email):
...
+        Preferred approach to avoid overloading the network.
+        :return: True or False
+        """
+        mailing_contact = self.search([("partner_id", "=", partner_id)], limit=1)
+        if mailing_contact:
+            try:
+                mailing_contact.action_update_to_mailchimp()
+            except:
+                return False
+        return True
```
ecino, line 95: Method `update_all_merge_fields_job` just below should use your new methods and launch a job for each contact instead of doing the whole loop. The whole purpose of the task was to split this job otherwise we will have difficulties to relaunch it when we have an error occurring at the middle of the sync.

ecino (changes requested):

mass_mailing_switzerland/models/mailchimp_lists.py
JonathanGuerne: The fetched partner are not given in clear to the function call. They are deducted from the response data by mailchimp in their code. reproducing such behavior would require a lot of code and be redundant.

The value of the `is_email_valid` field is set in the mailchimp function mentioned above. So adding the modification in place wouldn't be an option. 

I see 2 options : 
- unlink mailcontact when communication is send so the only existing mail_contact with invalid emails are the new ones
- setting up an onchange function in the partner to track is_email_valid value

ecino: I think the last option is the best, but not using onchange which is reserved for user interface change. The proper way is to inherit the write method and check there if is_email_valid value is changed.

mass_mailing_switzerland/models/mass_mailing_contact.py
```diff
@@ -42,6 +42,56 @@ def get_partner(self, email):
...
+        """Update one contact merge fields.
+        Preferred approach to avoid overloading the network.
+        :return: True or False
+        """
+        mailing_contact = self.search([("partner_id", "=", partner_id)], limit=1)
+        if mailing_contact:
+            try:
+                mailing_contact.action_update_to_mailchimp()
+            except:
+                return False
```
ecino: Since now we update only one contact at a time, I doubt it's useful to catch the error. We better remove the `try` block and raise the error. The job will be put in failed state and we will be able to see the stack trace of the error, which be more useful than having a "done" job that returned false with no info.

mass_mailing_switzerland/models/mass_mailing_contact.py
```diff
@@ -79,8 +122,33 @@ def action_update_to_mailchimp(self):
...
+            # if previous write failed reference to mailchimp member will be lost. error 404
+            try:
+                out = out and super(MassMailingContact,
+                                    contact_to_update.with_context(lang="en_US")).action_update_to_mailchimp()
+            except Exception as e:
+                # if not contact were fond. means an error occur at last write().
+                # Email field in odoo and mailchimp are now different.
+                # solution : we remove previous link to mailchimp and export the contact with new mail
+                if e.args[0] and literal_eval(e.args[0])['status'] == 404:
+                    available_mailchimp_lists = self.env['mailchimp.lists'].search([])
```
ecino, line 141:
```suggestion
                    self.env.clear()
                    available_mailchimp_lists = self.env['mailchimp.lists'].search([])
```

mass_mailing_switzerland/models/mass_mailing_contact.py
```diff
@@ -79,8 +122,33 @@ def action_update_to_mailchimp(self):
...
+                    lists = available_mailchimp_lists.mapped('odoo_list_id').ids
+
+                    contact_to_update.subscription_list_ids.filtered(
+                        lambda x: x.list_id.id in lists).write({"mailchimp_id": False})
+                # raise exception if it's any other type
+                else:
+                    self.env.clear()
```
ecino: The `clear` is not at the right place. No need to clear if we simply raise the error.
```suggestion
```

mass_mailing_switzerland/models/res_partner.py
```diff
@@ -165,8 +165,15 @@ def write(self, vals):
...
-                .with_delay().update_all_merge_fields_job(update_partner_ids)
+            for partner_id in update_partner_ids:
+                queue_job = self.env["queue.job"].search([
+                    ("channel", "=", "root.mass_mailing_switzerland.update_partner_mailchimp"),
+                    ("state", "!=", "done")
+                ])
+                if len(queue_job) and partner_id in queue_job.mapped("args"):
+                    continue
+                self.env["mail.mass_mailing.contact"] \
+                    .with_delay().update_partner_merge_fields_job(partner_id)
```
ecino: I think this logic could be moved inside the job function `update_partner_merge_fields_job` so that here we wouldn't need to change the code (plus this would be done each time the job is called instead of just here).

mass_mailing_switzerland/models/res_partner_category.py
```diff
@@ -17,14 +17,22 @@ class ResPartnerCategory(models.Model):
...
-            .with_delay().update_all_merge_fields_job(self.mapped("partner_ids").ids)
+            for partner_id in old_partners ^ new_partners:
+                queue_job = self.env["queue.job"].search([
+                    ("channel", "=", "root.mass_mailing_switzerland.update_partner_mailchimp"),
+                    ("state", "!=", "done"),
+                ])
+                if len(queue_job) and partner_id in queue_job.mapped("args"):
+                    continue
+                self.env["mail.mass_mailing.contact"] \
+                    .with_delay().update_partner_merge_fields_job(partner_id)
```
ecino: Here we could as well simply call the job functions on the right recordset (`old_partners ^ new_partners`) if we move the logic of searching for existing jobs inside the `update_partner_merge_fields_job` function.

ecino (changes requested):

mass_mailing_switzerland/models/res_partner.py
JonathanGuerne: :+1: 
I will use `update_all_merge_fields_job` since it can be call with a list of partners and will launch a job for each one of them. It also take care of the delay call

ecino (approved):

PR #1289, opened 2021-03-09
CO-3027 change regex to avoid unwanted behavior

JonathanGuerne:
Certain correspondent_id (store on the bvr ref) tricked the regex into matching on the wrong part of the string. I've tested the new regex with a couple more bvr to assess good overall behavior.

### DIFF
--- a/sponsorship_switzerland/models/completion_rules.py
+++ b/sponsorship_switzerland/models/completion_rules.py
@@ -288,8 +288,7 @@ class StatementCompletionRule(models.Model):
                 "|",
                 ("partner_id", "=", partner.id),
                 ("correspondent_id", "=", partner.id),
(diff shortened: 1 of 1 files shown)


sponsorship_switzerland/models/completion_rules.py
```diff
@@ -289,6 +289,7 @@ def _generate_invoice(self, stmts_vals, st_line, partner, ref_index):
                 ("partner_id", "=", partner.id),
                 ("correspondent_id", "=", partner.id),
                 ("state", "!=", "draft"),
+                ("state", "!=", "terminated"),
```
coup1234:
```suggestion
                ("state", "not in", ("terminated","draft"))
```

ecino (approved):

PR #1309, opened 2021-04-07
CO-3488 add form to make donation to available funds

JonathanGuerne:
- add a `website_published` field to `product.product` 
- available fund to donate to are display in a grid of card available from my donations page

### DIFF
--- a/website_compassion/controllers/my_account.py
+++ b/website_compassion/controllers/my_account.py
@@ -417,18 +417,17 @@ class MyAccountController(PaymentFormController):
 
         selected_fund_id = available_funds.browse(int(selected_fund_id))
 
-        values = kwargs.copy()
         # This allows the translation to still work on the page
-        values.pop("edit_translations", False)
-        values.update({
+        kwargs.pop("edit_translations", False)
+        kwargs.update({
             "form_model_key": "cms.form.fund.donation",
             "selected_fund": selected_fund_id,
         })
 
-        donation_form = self.get_form("res.partner", partner.id, **values)
+        donation_form = self.get_form("res.partner", partner.id, **kwargs)
         donation_form.form_process()
 
-        values.update({
+        kwargs.update({
             "form": donation_form,
         })
 
@@ -436,10 +435,10 @@ class MyAccountController(PaymentFormController):
         if donation_form.form_success:
             return werkzeug.utils.redirect(donation_form.form_next_url(), code=303)
 
-        values.update(self._prepare_portal_layout_values())
+        kwargs.update(self._prepare_portal_layout_values())
 
         result = request.render(
-            "website_compassion.new_donation_to_fund_template", values
+            "website_compassion.new_donation_to_fund_template", kwargs
         )
 
         return self._form_redirect(result, full_page=True)
@@ -448,9 +447,8 @@ class MyAccountController(PaymentFormController):
     def new_donation_validation(self, invoice_id=None, **kwargs):
 
         payment = kwargs.get("payment")
-        values = kwargs.copy()
 
-        values.update(self._prepare_portal_layout_values())
+        kwargs.update(self._prepare_portal_layout_values())
 
         try:
             invoice = request.env["account.invoice"].sudo().browse(int(invoice_id))
@@ -461,12 +459,12 @@ class MyAccountController(PaymentFormController):
 
         if transaction.state != "done" or payment == "error":
             result = request.render(
-                "website_compassion.donation_failure", values
+                "website_compassion.donation_failure", kwargs
             )
             return self._form_redirect(result, full_page=True)
 
         result = request.render(
-            "website_compassion.donation_success", values
+            "website_compassion.donation_success", kwargs
         )
         return self._form_redirect(result, full_page=True)
 
--- a/website_compassion/controllers/my_account.py
+++ b/website_compassion/controllers/my_account.py
@@ -401,18 +401,16 @@ class MyAccountController(PaymentFormController):
             "funds": available_funds
         })
 
-        result = request.render(
+        return request.render(
             "website_compassion.my_new_donation_template", values
         )
-
-        return self._form_redirect(result, full_page=True)
 
     @route("/my/new/donation/fund/<int:selected_fund_id>", type="http", auth="user", website=True)
     def make_new_donation_to_fund(self, selected_fund_id=None, **kwargs):
         partner = request.env.user.partner_id
         available_funds = request.env["product.product"].search([("website_published", "=", True)])
 
-        if not selected_fund_id or int(selected_fund_id) not in available_funds.mapped("id"):
+        if not selected_fund_id or int(selected_fund_id) not in available_funds.ids:
             return request.redirect("/my/new/donation")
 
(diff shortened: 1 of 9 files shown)


website_compassion/forms/fund_donation_form.py
```diff
@@ -0,0 +1,106 @@
...
+        return f"/my/new/donation/validate/{self.invoice_id.id}?payment=error"
+
+    @property
+    def form_description(self):
+        return self.fund.description
+
+    @property
+    def form_title(self):
+        if self.fund:
+            return _("Donation for ") + self.fund.name
```
coup1234, line 50: Check if this line is called/work

website_compassion/models/product_template.py
```diff
@@ -0,0 +1,22 @@
...
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+
+import logging
+
+from odoo import models, api, fields
+
+logger = logging.getLogger(__name__)
```
coup1234: Is the logger used ?

website_compassion/controllers/my_account.py
```diff
@@ -385,6 +386,90 @@ def my_donations(self, form_id=None, **kw):
...
+    def make_new_donation_to_fund(self, selected_fund_id=None, **kwargs):
+        partner = request.env.user.partner_id
+        available_funds = request.env["product.product"].search([("website_published", "=", True)])
+
+        if not selected_fund_id or int(selected_fund_id) not in available_funds.mapped("id"):
+            return request.redirect("/my/new/donation")
+
+        selected_fund_id = available_funds.browse(int(selected_fund_id))
+
+        values = kwargs.copy()
```
coup1234: I think in other parts of the code no copy is made. Is it important here ?

website_compassion/controllers/my_account.py
```diff
@@ -385,6 +386,90 @@ def my_donations(self, form_id=None, **kw):
...
+            "website_compassion.new_donation_to_fund_template", values
+        )
+
+        return self._form_redirect(result, full_page=True)
+
+    @route("/my/new/donation/validate/<int:invoice_id>", type="http", auth="user", website=True)
+    def new_donation_validation(self, invoice_id=None, **kwargs):
+
+        payment = kwargs.get("payment")
+        values = kwargs.copy()
```
coup1234: I think in other parts of the code no copy is made. Is it important here ?

website_compassion/controllers/my_account.py
```diff
@@ -385,6 +386,90 @@ def my_donations(self, form_id=None, **kw):
...
+
+        values.update(self._prepare_portal_layout_values())
+
+        try:
+            invoice = request.env["account.invoice"].sudo().browse(int(invoice_id))
+            invoice.exists().ensure_one()
+            transaction = invoice.get_portal_last_transaction()
+        except ValueError:
+            transaction = request.env["payment.transaction"]
```
coup1234, line 457: Maybe it would be better, if possible, to do an 'if statement' here instead of a try, except. Or raise an error if it is an unexpected behaviour.

website_compassion/forms/fund_donation_form.py
```diff
@@ -0,0 +1,106 @@
...
+            return "amount", _("Please set an amount")
+        return 0, 0
+
+    @property
+    def _payment_success_redirect(self):
+        return f"/my/new/donation/validate/{self.invoice_id.id}?payment=success"
+
+    @property
+    def _payment_error_redirect(self):
+        return f"/my/new/donation/validate/{self.invoice_id.id}?payment=error"
```
coup1234: Maybe add a TODO to check what we can do because I think this function is never called. There is a page with the list of cancelled invoice instead.

website_compassion/forms/fund_donation_form.py
JonathanGuerne: after verification it does work properly

website_compassion/controllers/my_account.py
JonathanGuerne: This code can also be found in other module. If getting the last transaction cause some issue it will raise a Value Error. To avoid further problem in the function logic we set transaction to an empty one. This will not be an issue because during the next step we check that its state is equal to "done" or we redirect to the error page

website_compassion/controllers/my_account.py
```diff
@@ -385,6 +386,88 @@ def my_donations(self, form_id=None, **kw):
...
+
+        values.update({
+            "funds": available_funds
+        })
+
+        result = request.render(
+            "website_compassion.my_new_donation_template", values
+        )
+
+        return self._form_redirect(result, full_page=True)
```
ecino:
```suggestion
        return request.render(
            "website_compassion.my_new_donation_template", values
        )
```

website_compassion/controllers/my_account.py
```diff
@@ -385,6 +386,88 @@ def my_donations(self, form_id=None, **kw):
...
+
+    @route("/my/new/donation/fund/<int:selected_fund_id>", type="http", auth="user", website=True)
+    def make_new_donation_to_fund(self, selected_fund_id=None, **kwargs):
+        partner = request.env.user.partner_id
+        available_funds = request.env["product.product"].search([("website_published", "=", True)])
+
+        if not selected_fund_id or int(selected_fund_id) not in available_funds.mapped("id"):
```
ecino:
```suggestion
        if not selected_fund_id or int(selected_fund_id) not in available_funds.ids:
```

website_compassion/forms/fund_donation_form.py
```diff
@@ -0,0 +1,107 @@
...
+
+    _display_type = "full"
+    _form_model = "res.partner"
+
+    amount = fields.Float(required=True)
+    fund = fields.Many2one("product.product")
+
+    _form_model_fields = [
+        "amount",
+    ]
```
ecino, line 26: This is wrong, it would mean the field "amount" is in the underlying model which here you set to "res.partner".
```suggestion
```

website_compassion/forms/fund_donation_form.py
```diff
@@ -0,0 +1,107 @@
...
+        return form
+
+    def generate_invoice(self):
+        fund = self.fund.sudo()
+        name = f"Donation for {fund.name}"
+        partner = self.main_object.sudo()
+        return self.env["account.invoice"].sudo().create(
+            {
+                "name": name,
+                "origin": name,
```
ecino: Maybe this will be more useful.
```suggestion
                "origin": "MyCompassion",
```

website_compassion/models/product_template.py
```diff
@@ -0,0 +1,20 @@
...
+import logging
+
+from odoo import models, api, fields
+
+
+class ProductProduct(models.Model):
+    _name = "product.template"
+    _inherit = "product.template"
+
+    website_published = fields.Boolean(index=True)
```
ecino: The correct way is to inherit from an abstract class already implemented for this purpose:
```suggestion
    _name = "product.template"
    _inherit = ["product.template", "website.published.mixin"]
```

ecino (changes requested):

PR #1310, opened 2021-04-07
CO-3595 add pagination and change/move some call to improve performance

JonathanGuerne:
- fix internal error on donations when user doesn't have any sponsorship
- add pagination
- improve loading performance by moving function call to backend


website_compassion/controllers/my_account.py
```diff
@@ -270,24 +272,43 @@ def my_child(self, state="active", child_id=None, **kwargs):
...
             ("type", "=", "out_invoice"),
             ("amount_total", "!=", 0),
-        ])
+        ]
+
+        sudo_invoice_env = request.env["account.invoice"].sudo()
+
+        all_invoice_count = sudo_invoice_env.search_count(invoice_search_criteria)
+
+        invoice_per_page = 30
```
coup1234: Suggestion: maybe the invoice_per_page can be a parameter of the function with default = 30

website_compassion/controllers/my_account.py
```diff
@@ -296,23 +317,25 @@ def my_donations(self, form_id=None, **kw):
             ("date_invoice", "<", fields.Date.to_string(in_one_month))
         ])
 
+        last_completed_tax_receipt = partner.last_completed_tax_receipt
```
coup1234, line 323: Maybe this init is not really needed as we only use it on time (line 406)

JonathanGuerne: Yes the call is actually only needed in the creation of the dictionary used in the return statement. But when called directly in the dictionary it led to a strange behavior where last_completed_tax year was computed not only for the current user but also for other people. Maybe it has something to do with the state of self changing during the dictionary preparation. either way calling it before, as it is done here, solved the issue

ecino (approved):

website_compassion/controllers/my_account.py
```diff
@@ -356,27 +381,31 @@ def my_donations(self, form_id=None, **kw):
...
             ("partner_id", "=", partner.id),
             ("state", "=", "paid"),
             ("type", "=", "out_invoice"),
             ("amount_total", "!=", 0),
-        ], limit=1, order="create_date asc")
+        ], limit=1, order="create_date asc").create_date
+
         current_year = datetime.today().year
-        first_year = first_invoice.create_date and first_invoice.create_date.year or current_year
+        first_year = create_date.year if create_date or current_year
```
ecino:
```suggestion
        first_year = create_date.year if create_date else current_year
```

PR #1314, opened 2021-04-07
CO-3608, CO-3606: Double pictures and view letter correction.

coup1234:
- Change the url link photo to prevent double
- Correct the parameter names when calling bs2_image.
- No get_required_param  when calling download/source


davidwul:
@ecino 
I have tested on the devel, corrections are effectives

PR #1318, opened 2021-04-09
CO-3609: - Replace the first, lastname field by name field in Coordin…

coup1234:
…ates form for company 
- sudo context when modify the partner

### DIFF
--- a/website_compassion/forms/partner_forms.py
+++ b/website_compassion/forms/partner_forms.py
@@ -60,11 +60,6 @@ class PartnerCoordinatesForm(models.AbstractModel):
         "email",
     ]
 
-    def _form_write(self, values):
-        """ Nothing to do on write, we handle everything in other methods.
-        """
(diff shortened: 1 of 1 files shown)


JonathanGuerne (approved):

website_compassion/forms/partner_forms.py
```diff
@@ -50,18 +52,34 @@ class PartnerCoordinatesForm(models.AbstractModel):
...
         "city",
         "country_id",
         "email",
     ]
 
+    def _form_write(self, values):
+        """ Nothing to do on write, we handle everything in other methods.
+        """
+        return self.main_object.sudo().write(values.copy())
+
```
ecino: Wrong comment, plus we don't want to write as sudo otherwise we lose the info that the sponsor made himself the changes. It's better to add missing security rules if there are any (we already added some in production).
```suggestion
```

ecino (changes requested):

PR #1319, opened 2021-04-09
CO-3603 load minimized template letter in my/letter

JonathanGuerne:
problem:
- full size image load ( often > 1 MB) when only a smaller version is used

solution:
- use web/image route to load a resize version of the image. keep width/height ratio. by default divided by 10

### DIFF
--- a/website_compassion/models/correspondence_template.py
+++ b/website_compassion/models/correspondence_template.py
@@ -15,10 +15,15 @@ class CorrespondenceTemplate(models.Model):
     website_published = fields.Boolean()
     template_image_url = fields.Char(compute="_compute_template_image_url")
 
-    def _compute_template_image_url(self):
+    def _compute_template_image_url(self, resize_factor=10):
         web_base_url = (self.env["ir.config_parameter"].sudo()
                         .get_param("web.external.url"))
+
+        def resize_dimension(template):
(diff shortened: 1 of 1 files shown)


website_compassion/models/correspondence_template.py
```diff
@@ -15,10 +15,15 @@ class CorrespondenceTemplate(models.Model):
...
     template_image_url = fields.Char(compute="_compute_template_image_url")
 
-    def _compute_template_image_url(self):
+    def _compute_template_image_url(self, resize_factor=10):
         web_base_url = (self.env["ir.config_parameter"].sudo()
                         .get_param("web.external.url"))
+
+        def resize_dimension(template):
+            return "x".join(list(map(lambda dim: str(round(dim / resize_factor)),
+                                     template.get_template_size(1))))
```
coup1234:
```suggestion
        def resize_dimension(template):
            dim = template.get_template_size(1/resize_factor)
            return f'{round(dim[0])}x{round(dim[1])}'
```

coup1234: Or maybe we can set resize_factor = 0.1 and put `template.get_template_size(resize_factor)`

ecino:
Corrected in e168cf61397fbdb3a6d5d3001cad46ba512f342b
Sorry for the work, in fact the /web/image route directly supports resizing.

PR #1321, opened 2021-04-15
CO-3531: - New field cancel_invoice, if true the invoice will be canc…

coup1234:
…el in the case where the payment transaction is cancelled
 - Add a cron that cancel payment.transaction if state = draft during 10 min (to be activated)


website_compassion/data/base_automation_data.xml
```diff
@@ -0,0 +1,13 @@
+<odoo>
+        <record id="automated_action" model="base.automation">
+            <field name="name">Cancel Draft Transaction after 10 min</field>
+            <field name="model_id" ref="model_payment_transaction"/>
+            <field name="trigger">on_time</field>
+            <field name="trg_date_range">1</field>
```
ecino: Every 1 minute seems agressive to me.
```suggestion
            <field name="trg_date_range">10</field>
```

website_compassion/models/invoice_line.py
```diff
@@ -6,9 +6,15 @@
...
 
 
 class AccountInvoiceLine(models.Model):
     _name = "account.invoice.line"
     _inherit = ["account.invoice.line", "translatable.model"]
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+    # If true, cancel the invoice if the linked payment transaction is cancelled
```
ecino: This string can be put in the `help` attribute of the field
```suggestion
    # If true, cancel the invoice if the linked payment transaction is cancelled
```

website_compassion/models/invoice_line.py
```diff
@@ -6,9 +6,15 @@
...
+    cancel_invoice = fields.Boolean(default=False)
```
ecino: I would prefer a more descriptive field_name. This can lead to confusion.
```suggestion
    auto_cancel_no_transaction = fields.Boolean(default=False)
```

website_compassion/models/payment_transaction.py
```diff
@@ -0,0 +1,19 @@
...
+#
+##############################################################################
+from odoo import models, fields
+
+
+class PaymentTransaction(models.Model):
+    _inherit = "payment.transaction"
+
+    def _set_transaction_cancel(self):
+        for invoice in self.invoice_ids.filtered(lambda a: a.cancel_invoice):
```
ecino: If the filter function is directly on a field, you can simplify and avoid using a lambda.
```suggestion
        for invoice in self.invoice_ids.filtered('auto_cancel_no_transaction'):
```

ecino (changes requested):

ecino (changes requested):
I think we should move all these changes in module `cms_form_compassion` so that the logic is also accessible to compassion-modules modules, especially `mobile_app_connector`.

Then you can remove the dependency added to `website_compassion` as the module `wordpress_connector` already depends on `cms_form_compassion`(and `mobile_app_connector` as well).

website_compassion/models/invoice_line.py
```diff
@@ -6,9 +6,11 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
-from odoo import models
+from odoo import models, fields
```
ecino:
```suggestion
from odoo import models
```

ecino (changes requested):

ecino (approved):

PR #1327, opened 2021-04-27
FIX issue on product assignation

JonathanGuerne:
- previous update had introduced matching product errors in production

- change regexp to match the expected bvr format (shown by David)
  - Tested on gift, sponsorship and fund


ecino:
I fast track this into production.

PR #1328, opened 2021-04-27
CO-3623 override new mailchimp method to launch job on update

JonathanGuerne:
:warning: **Update mailchimp module with new modifications before merging this PR**

- override `process_mailchimp_update` to change partner update to mailchimp behavior
- launch already created job `update_all_merge_fields` for specified partners

This will ensure all update to mailchimp append in a separate thread (`with_delay`) thus avoiding issue on immediate update.

### DIFF
--- a/mass_mailing_switzerland/models/res_partner.py
+++ b/mass_mailing_switzerland/models/res_partner.py
@@ -144,7 +144,8 @@ class ResPartner(models.Model):
             vals["date_opt_out"] = False
         base_mailchimp_fields = [
             "email", "lastname", "firstname", "number_sponsorships", "opt_out"]
-        update_partner_ids = []
+
+        update_partner_ids = self.env["res.partner"]
         for contact in self.filtered(lambda c: c.mailing_contact_ids):
             if "opt_out" in vals and vals["opt_out"] != contact.opt_out:
                 contact.mailing_contact_ids.mapped("subscription_list_ids") \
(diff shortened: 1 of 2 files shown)


ecino:
I think this would already handled by https://github.com/CompassionCH/compassion-switzerland/pull/1349.
The changes there pushes the changes in the mailing_contact model and from there we launch a job. Can you check? In that case we could close this PR.

JonathanGuerne:
It is a slightly different approach than what could be found in the mentioned commit. 
The base automation while working was very load heavy because it is launch after each partner update. The new approach is to override a function that is called specifically on partner update to mailchimp. 

But I forgot the mailchimp part of this task (the code is ready but I am not able to find any PR...sorry).

ecino:
Oh I forgot the base automation, but couldn't we just remove it and it should work with #1349 or am I missing a piece? I think any update on the partner would trigger a write on the mass_mailing_contact that would itself call an asyncrhonous job.

But if you think we miss something, then we should adapt your changes to #1349 because it changes the code a lot.

JonathanGuerne:
You're right 

we modified the `res.partner` write function inside the `mass.mailing.switzerland` module to trigger an update job when selected fields are modified. So it will be ok to leave the code as is.

We only have to be careful about the code in the `mailchimp` module itself. If you look at the `res.partner` inside the `mailchimp` module you'll see that it also update the partner changes to mailchimp and, as we realized, it can cause issue because it happen immediately. 

If we make sure that the mailchimp module is not performing update by itself we could avoid those issue

ecino:
Will be handled in #1349

PR #1356, opened 2021-06-15
CO-3629 Zoom registrations and reminders

ecino:
- Add web page for confirming presence
- Automatically add new sponsors as participants when New Dossier is sent
- Add different communications for reminders

### DIFF
--- a/partner_communication_switzerland/models/res_partner_zoom_attendee.py
+++ b/partner_communication_switzerland/models/res_partner_zoom_attendee.py
@@ -1,0 +1,102 @@
+##############################################################################
+#
+#    Copyright (C) 2016 Compassion CH (http://www.compassion.ch)
+#    Releasing children from poverty in Jesus' name
+#    @author: Emanuel Cino <ecino@compassion.ch>
+#
+#    The licence is in the file __manifest__.py
+#
+##############################################################################
+from datetime import datetime
+
+from dateutil.relativedelta import relativedelta
+
+from odoo import models, fields
+
+
+COLOR_MAPPING = {
+    "invited": 0,
+    "confirmed": 10,
+    "declined": 1,
+    "attended": 11
+}
+
+
+class ZoomAttendee(models.Model):
+    _name = "res.partner.zoom.attendee"
+    _inherit = ["mail.thread", "mail.activity.mixin"]
+    _description = "Zoom Attendee"
+    _rec_name = "partner_id"
+    _order = "id desc"
+
+    partner_id = fields.Many2one("res.partner", "Partner", required=True, index=True)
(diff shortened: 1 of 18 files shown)


ecino:
Not yet fully tested but can still be reviewed.

partner_communication_switzerland/models/res_partner_zoom_attendee.py
```diff
@@ -0,0 +1,102 @@
...
+         "This partner is already invited to this zoom session!")
+    ]
+
+    def _compute_color(self):
+        for attendee in self:
+            attendee.color = COLOR_MAPPING.get(attendee.state)
+
+    def inform_about_next_session(self):
+        for attendee in self:
+            next_zoom = attendee.zoom_session_id.get_next_session()
```
JonathanGuerne, line 55: is lang context implied by attendee or might we have to specify it ?

partner_communication_switzerland/views/notification_settings_view.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+    Copyright (C) 2021 Compassion (http://www.compassion.ch)
+    @author Emanuel Cino <ecino@compassion.ch>
+    The licence is in the file __manifest__.py
+-->
+<odoo>
+    <!-- Configuration View -->
+    <record id="view_notification_settings_partner_compassion_form" model="ir.ui.view">
+        <field name="name">staff.notification.settings.muksathlon.form</field>
```
JonathanGuerne: Is the muskathlon mention voluntary ?

partner_communication_switzerland/models/res_partner_zoom_attendee.py
ecino: Maybe the method name is not very good because it's not at this time that we inform him. We just take note that he wants to be informed. Here we don't need to care about his language.

partner_communication_switzerland/views/notification_settings_view.xml
ecino: it's a bad copy/paste, I will fix it. It doesn't make any problem however.

partner_communication_switzerland/models/res_partner_zoom_attendee.py
JonathanGuerne: I'm surprised because you used a context before : https://github.com/CompassionCH/compassion-switzerland/pull/1356/files#diff-56fb2534043f2802633676aad3d448b0aed27b7477958b255247be02335d34a5R617

and the function that look over available zoom meeting use lang as a search field

JonathanGuerne (approved):

PR #1357, opened 2021-06-16
CO-3660 based on new display type option display correspondent letters idf needed

JonathanGuerne:
- change context to sudo() if needed
- look for correspondent letters


ecino (changes requested):
The change is good but I think we miss the fact that the user should be also able to write to the child. Will he see the child under the "write a letter" a page? It should be possible.

JonathanGuerne:
> The change is good but I think we miss the fact that the user should be also able to write to the child. Will he see the child under the "write a letter" a page? It should be possible.

Yes the payer only can see the child once the display option has been changed. I've just checked

ecino (approved):

PR #1360, opened 2021-06-22
CO-3629 fix issues with form and is_first_sponsorship

JonathanGuerne:
- change the way the main_object is created. otherwise it returned an empty zoom attendee
- move ìs_first_sponsorship` computation to happened before the `_new_dossier` function
- change param references that caused issues


ecino (approved):
Thanks for those corrections :+1:

PR #1367, opened 2021-06-25
CO-3625 Batch printing communications

ecino:
Force printing in duplex when pair page count, otherwise single-sided

Depends on https://github.com/CompassionCH/compassion-modules/pull/1541

### DIFF
--- a/partner_communication_switzerland/models/partner_communication.py
+++ b/partner_communication_switzerland/models/partner_communication.py
@@ -52,6 +52,19 @@ class PartnerCommunication(models.Model):
         readonly=False,
     )
 
+    def print_letter(self, print_name, **print_options):
+        """
+        Adds duplex printing option for Konica Minolta depending on page count.
+        """
+        if len(self) > 1:
+            page_counts = list(set(self.mapped("pdf_page_count")))
+            # Duplex if all documents have a pair page count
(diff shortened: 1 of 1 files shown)


partner_communication_switzerland/models/partner_communication.py
```diff
@@ -52,6 +52,19 @@ class PartnerCommunication(models.Model):
...
+        """
+        Adds duplex printing option for Konica Minolta depending on page count.
+        """
+        if len(self) > 1:
+            page_counts = list(set(self.mapped("pdf_page_count")))
+            # Duplex if all documents have a pair page count
+            if len(page_counts) == 1 and page_counts[0] % 2 == 0:
+                print_options["KMDuplex"] = "2Sided"
+            else:
+                print_options["KMDuplex"] = "1Sided"
```
JonathanGuerne: Wouldn't that mean that for print option to be set to "2sided" all communication should be of the same length ?

The following should work for different len of communication as long as they are all of even length
```suggestion
            sided_option = "2sided"
            for p_count in page_counts: 
                if (p_count % 2 != 0):
                    sided_option = "1Sided"
                    break
            print_options["KMDuplex"] = sided_option

```

PR #1373, opened 2021-07-01
CO-3645 compress profile picture on mycompassion

JonathanGuerne:


ecino (changes requested):
If I don't mistake we have now two times this image_compression tool both in `crowdfunding` and in `website`. I think you could remove the one in `crowdfunding`and replace the imports with `from odoo.addons.website_compassion.tools.image_compression import compress_big_images` in order to avoid duplicate code to maintain.

ecino (approved):

PR #1374, opened 2021-07-02
CO-3688 a specific logic on mail delivery failure for letter and onboarding

JonathanGuerne:
:warning: for now those communication were sent either 2 or 3 days after the initial communication if it wasn't open. This logic is still in place and could, in some case, cause some duplicate communication

### DIFF
--- a/partner_communication_switzerland/models/mail_tracking_event.py
+++ b/partner_communication_switzerland/models/mail_tracking_event.py
@@ -100,15 +100,55 @@ class MailTrackingEvent(models.Model):
     def _invalid_email(self, tracking_email):
         """
         Sends invalid e-mail communication.
+        - if the invalid event came from an email linked to a correspondence send child letter unread
+        - if the invalid event came from an email link to the onboarding confirmation send unread confirmation
         """
+        
+        # invalid config to send to the partner
         invalid_comm = self.env.ref("partner_communication_switzerland.wrong_email")
+        b2s_email_not_read = self.env.ref("partner_communication_switzerland.child_letter_unread")
+        onboarding_welcome_email_not_read = self.env.ref(
+            "partner_communication_switzerland.config_onboarding_sponsorship_confirmation_not_read")
+
+        _config_id = invalid_comm
+        _object_ids = None
+
+        # correspondence mail should return b2s email not read
+        correspondence_email = tracking_email.filtered(
+            lambda x: x.mail_id.model == "partner.communication.job" and
+                      self.env["partner.communication.job"].browse(
+                          x.mail_id.res_id).model == "correspondence"
+        )
+
+        # other special case for onboarding welcome not read
+        onboarding_welcome_config = self.env.ref(
+            "partner_communication_switzerland.config_onboarding_sponsorship_confirmation")
+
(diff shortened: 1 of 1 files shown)


ecino (changes requested):
There were already two action rules handling the case for B2S unread and Welcome unread (sorry I only remember now), but they were never called I think because we didn't log the events (the problem we saw that the tracking event was not saved):

b2s : https://github.com/CompassionCH/compassion-switzerland/blob/f9f476ba8bd8adec8caa0abd7ab79fac01e739dc/partner_communication_switzerland/models/correspondence.py#L205
onboarding: https://github.com/CompassionCH/compassion-switzerland/blob/b1bcff99873a5932a45fbff3e50918ea7f04bef7/partner_communication_switzerland/data/onboarding_process.xml#L312

This makes this code redundant I think, but the change should maybe avoid sending any communication in these two cases that are already handled by the action rules. Or we can remove the action rules if this code is more accurately called than the action rules (verify that the triggers match). Filter called to determine if email is read or not: https://github.com/CompassionCH/compassion-switzerland/blob/b1bcff99873a5932a45fbff3e50918ea7f04bef7/partner_communication_switzerland/models/partner_communication.py#L83

JonathanGuerne:
For what I've seen I think avoiding to send to communication in  the ` _invalid_mail` function would be the better idea. `filter_not_read` is a more complete checkup of the email status even if it doesn't append as quickly as the other function (checkup after 2/3 days vs trigger by mandrill event)

ecino:
Ok, it seems good to me. Should we completely replace this method then by another action rule dealing with all communications that are neither B2S letter nor Onboarding welcome?

JonathanGuerne:
I don't know, that seems like an heavy action rule to setup as It should be looking for all outgoing mail. And, again, we would loss the direct generation of the invalid mail communication

ecino:
Ok good, so let's keep it like that. I'm waiting for your changes before merging.

JonathanGuerne:
Just to clarify only the communication will not be sent. The partner's invalid_mail field will be updated. This would append anyway if in between the "special communication" and our rule based checkup an other email could not be sent properly

partner_communication_switzerland/models/mail_tracking_event.py
```diff
@@ -100,21 +100,53 @@ def process_reject(self, tracking_email, metadata):
...
+        # return to avoid duplication
+        if correspondence_email or onboarding_welcome_email:
+            return
+
         partner_id = tracking_email.partner_id.id
         if partner_id:
             self.env["partner.communication.job"].create(
                 {
-                    "config_id": invalid_comm.id,
+                    "config_id": invalid_comm,
```
ecino: The previous version was correct. `invalid_comm` is a record and in the values dictionary we want to put the `id` of the record.

ecino (changes requested):

partner_communication_switzerland/models/mail_tracking_event.py
JonathanGuerne: Sorry for that... thank you

ecino (approved):

PR #1375, opened 2021-07-02
CO-3682 overwrite reset_function to make it case insensitive for email matching

JonathanGuerne:


website_compassion/models/res_user.py
```diff
@@ -0,0 +1,30 @@
...
+            and reset their password
+
+            case insensitive for email matching
+        """
+        users = self.search([('login', '=', login)])
+        if not users:
+            users = self.search([('email', '=ilike', login)])
+        if len(users) != 1:
+            raise Exception(_('Reset password: invalid username or email'))
+        return users.action_reset_password()
```
ecino: I would suggest to still use the parent method for maintenance reasons:
```suggestion
        try:
            return super().reset_password(login)
        except:
            users = self.search([('email', '=ilike', login)])
            if len(users) != 1:
                raise
            return users.action_reset_password()
```

ecino (changes requested):

ecino (approved):

PR #1376, opened 2021-07-06
CO-3564 prepare card report with signature

JonathanGuerne:


ecino (changes requested):
I would suggest this instead (after testing the result): This gives a better layout as the short_signature is on one line but we don't have good space so we need on two lines.
```
<div id="signature">
                        <div style="position: relative; top: -15mm">
                          <t t-set="sign_user_name">sarah maria</t>
                          <t t-set="sign_user" t-value="o.env['res.users'].search([('name', 'ilike', sign_user_name),('share','=',False)], limit=1)"/>
                          <img t-if="sign_user.digital_signature" t-attf-src="/web/image/res.users/#{sign_user.id}/digital_signature?width=200" style="position: relative; left: -10mm;"/>
                          <p>
                            <span style="font-weight: bold">Sarah Maria Herrendorff</span>
                            <br/>
                            <span>and the team from Compassion</span>
                          </p>
                        </div>
                    </div>
```

report_compassion/report/anniversary_card.xml
```diff
@@ -121,11 +121,19 @@
                         <div id="body">
                             <span t-raw="o.body_html"/>
                         </div>
+                        <t t-set="sign_user_name">sarah maria</t>
+                        <t t-set="sign_user" t-value="o.env['res.users'].search([('name', 'ilike', sign_user_name)], limit=1)"/>
```
ecino: This can be removed
```suggestion
```

ecino (changes requested):

ecino (approved):

PR #1377, opened 2021-07-13
CO-3678 use send_gifts_to to determine which partner should be loaded

JonathanGuerne:


wordpress_connector/models/match_partner_wp.py
```diff
@@ -79,5 +79,7 @@ def _match_get_rules_order(self):
     @api.model
     def _match_rule_child_id(self, partner_obj, infos, options=None):
         # if a keyerror is raise it is handled as "no child found go to next rule"
-        child_id = infos.pop("child_id")
-        return self.env["compassion.child"].search([("local_id", "like", child_id)]).sponsor_id
+        child_local_id = infos.pop("child_id")
+        child = self.env["compassion.child"].search([("local_id", "like", child_local_id)])
+        sponsorship = self.env["recurring.contract"].search([("child_id", "=", child.id)])
```
ecino: This would ensure to only take the latest sponsorship.
```suggestion
        sponsorship = self.env["recurring.contract"].search([("child_id", "=", child.id)], limit=1)
```

ecino (changes requested):
The change is good but I'm not sure it resolves the issue. I think the issue occurs later because the invoice generated doesn't always contain the sponsorship (field contract_id on account.invoice.line).

JonathanGuerne:
The contract_id could only be empty if the partner that the invoice would be sent to had no link with the referred child (we never interfere with the value of child_id). This is typically the case when the matching process failed and a new partner was created. 
In that case solving the matching issue would solve the missing contract information in the invoice as well.    

Although I would be wrong if we have any recent case of this issue ( more recent then the original PR changes). Do you have any information about that ?

ecino:
Ok, looks good.

ecino (approved):

PR #1384, opened 2021-07-22
CO-3688 create a new report based on submitted design

JonathanGuerne:
The newly created paper format requires the --disable-smart-shrinking option to generate the pdf properly. This could be added in code but would require to add an oca dependency to the compassion_report manifest

Montserrat font is not yet available in odoo. it should be provided preferably in otf format

The background drawing are also missing

### DIFF
--- a/partner_communication_switzerland/models/partner_communication.py
+++ b/partner_communication_switzerland/models/partner_communication.py
@@ -440,6 +440,22 @@ class PartnerCommunication(models.Model):
         )
         return {_("child dossier.pdf"): [report_name, pdf]}
 
+    def get_end_sponsorship_certificate(self):
+        self.ensure_one()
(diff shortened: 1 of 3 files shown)


ecino:
What is the dependency? It's probably a module we use (and maybe wrote as well), so I think it's not a problem to add it.

We should ask Olivier for the font or switch to Miller (I think we cannot push the Montserrat otf on Github because it's a paid font). I don't know what solution we can have for this.

JonathanGuerne:
It use the model `report.paperformat.parameter` which is located in the oca module `report_wkhtmltopdf_param`

ecino:
We use it already in production, I don't see an issue by adding the dependency since we are here on our swiss modules.

ecino (approved):

PR #1386, opened 2021-07-23
CO-3677 remove 'connect w/ google' button when not on base.url

JonathanGuerne:
The idea was to assess whether allowing user to sign in w/ google was easy to setup or not and implement changes accordingly. 

What I've found is that theoretically, at least for odoo 14.0, existing user could sign in trough the google button if they reset their password first [source](https://www.odoo.com/documentation/14.0/applications/general/auth/google.html). This should be tested for odoo 12.0.

If it should not work then this commit provide a simple  way to hide the google login button if the url used is not the base one.

### DIFF
--- a/website_compassion/__manifest__.py
+++ b/website_compassion/__manifest__.py
@@ -53,6 +53,7 @@
         "theme_compassion",
         "partner_communication_switzerland",
         "mobile_app_connector",
+        "auth_oauth"
     ],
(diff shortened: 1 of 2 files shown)


ecino:
Based on Andre's testing, it looks like it works fine even for external accounts. But Stephane will have to configure that. It means that we don't need to hide the button.

PR #1387, opened 2021-07-23
CO-3704 & CO-3664 Add QRcode to child packs and birthdate to childpack mini

JCompassion:


ecino (commented):
Changes look good, but I saw in the email exchange that the QR code should appear on all variants of the childpacks. Can you already add them or do we miss something for that ?

JCompassion:
It is already implemented in _report_compassion/models/child.py_. I edited the childpack_small and (used for small & full) as well as the childpack_mini.

ecino (approved):
All good then, thanks!

PR #1394, opened 2021-08-02
CO-3711 new donors onboarding process

JonathanGuerne:
- create missing communication
  - in german and french
- setup base automation for step 3 and 4
  - use change made by https://github.com/CompassionCH/compassion-modules/pull/1555
- filter partner
- save start date on new donors onboarding
  - can be use to stop the process
- update mass mailing smart tag
- send notification to "new donors user" when someone opt out


partner_communication_switzerland/models/res_partner.py
```diff
@@ -157,27 +161,27 @@ def _compute_new_donor(self):
...
+                                                 "only" in partner.letter_delivery_preference
+                                                 or partner.letter_delivery_preference == "none"
+                                         ) and (
+                                                 "only" in partner.photo_delivery_preference
+                                                 or partner.photo_delivery_preference == "none"
+                                         ) and (
+                                                 "only" in partner.thankyou_preference
+                                                 or partner.thankyou_preference == "none"
+                                         ) and partner.tax_certificate != "paper" and partner.nbmag in (
+                                             "email", "no_mag")
```
ecino: The identation was better before :)

partner_communication_switzerland/models/res_partner.py
```diff
@@ -218,10 +222,33 @@ def _compute_last_completed_tax_receipt(self):
...
+                "mail.mail_activity_data_email",
+                summary=_("Someone opt out of the new donor onboarding process"),
+                note=_("{} just opt out of the onboarding process for new donors.".
+                       format(partner.name)
+                       ),
+                user_id=new_donors_user
+            )
+
+        return res
+
```
ecino: I don't like this process, what is it supposed to do? Why do we schedule an activity here, isn't it something that can be handled automatically? What's the goal of this activity?

partner_communication_switzerland/models/res_partner.py
```diff
@@ -359,3 +386,23 @@ def action_reset_password(self):
             "res_id": comm.id,
             "target": "current",
         }
+
+    @api.multi
+    def filter_onboarding_new_donors(self):
+        return self.filtered(lambda p: p.is_new_donor and not p.is_church)
```
ecino: We should also make sure it's not already a sponsor

partner_communication_switzerland/models/res_partner.py
JonathanGuerne: We've been asked to inform a designated compassion staff once a new donor opt out of the process. This user can be set through settings. I specifically wanted to inform the person if the date was set but the removed.

partner_communication_switzerland/models/res_partner.py
JonathanGuerne: oops I'll look into it ^^'

JonathanGuerne:
I used `sponsorships_ids` to check if the partner has ever been a sponsor 
tell me if I need to change it for "is currently a sponsor"

ecino:
> I used `sponsorships_ids` to check if the partner has ever been a sponsor
> tell me if I need to change it for "is currently a sponsor"

I seems fine to me :+1:

JonathanGuerne:
I took the liberty to remove the translation as it seemed to cause a lot of conflict issue 
I'll add them in production once ready

partner_communication_switzerland/models/res_partner.py
```diff
@@ -383,3 +385,4 @@ def start_new_donors_onboarding(self):
                 })
 
             partner.onboarding_new_donor_start_date = fields.Date.today()
+            partner.onboarding_new_donor_hash = hashlib.md5(f"{partner.id}".encode()).hexdigest()
```
ecino: Will this always produce the same hash for a given partner id? Shouldn't we use a random hash like the uuid library provides?

JonathanGuerne: You're right the hash would be the same every time (for the same email address)
I can change it

ecino (approved):

PR #1395, opened 2021-08-09
CO-3727 migration to prevent duplicate ref

JonathanGuerne:


partner_compassion/migrations/12.0.1.1.4/post-migration.py
```diff
@@ -0,0 +1,37 @@
...
+def migrate(env, installed_version):
+    if not installed_version:
+        return
+
+    def get_partners_to_change():
+        all_companies_refs = env["res.partner"].search([("is_company", "=", True)]).mapped("ref")
+        return env["res.partner"].search([("is_company", "=", False), ("ref", "in", all_companies_refs)])
+
+    partner_to_change = get_partners_to_change()
+    sponsor_to_change = partner_to_change.filtered(lambda p: any(p.mapped("sponsorship_ids.is_active")))
```
ecino: I wonder if this works as well?
```suggestion
    sponsor_to_change = partner_to_change.filtered("sponsorship_ids.is_active")
```

ecino (changes requested):
- Did you test the migration? How many manual migration did you encounter?
- Could you as well remove the duplicate sequence partner.ref in the migration?

JonathanGuerne:
I tested it on my local database (which is about 2 month old now). I ended up with < 10 case that needed a manual inspection. 
Strange thing was that concerned partner had no parent_id. It seems to be an issue with the duplicated ref entry.

partner_compassion/migrations/12.0.1.1.4/post-migration.py
JonathanGuerne: it does !

JonathanGuerne:
As asked I've remove the unwanted `ir.sequence`. It required me to alter the `ir_model_data` table

There are 2 entries that match the following condition : 
```sql 
model = 'ir.sequence' and name = 'sequence_partner_ref'
```
Each entry point to a different module. 

If we don't alter this table removing the unwanted sequence will trigger it's recreation. 

I chose to change the `res_id` of one of the entry to copy the other one (base on which one seemed to be used by our system).

ecino (approved):

PR #1396, opened 2021-08-09
CO-3730 change utm.campaign amount and linked view

JonathanGuerne:


mass_mailing_switzerland/models/analytic_account.py
```diff
@@ -11,7 +11,12 @@
...
-
     _inherit = "account.analytic.account"
 
     campaign_id = fields.Many2one("utm.campaign", readonly=False)
+
+
+class AnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+
+    state = fields.Selection(related='move_id.invoice_id.state', readonly=False)
```
ecino: Why is this not readonly? I think it would be better to let it readonly.

mass_mailing_switzerland/models/utm.py
```diff
@@ -204,6 +204,31 @@ class UtmCampaign(models.Model):
         compute="_compute_click_count", store=True, readonly=True
     )
 
+    def open_invoices(self):
```
ecino: This would be a more appropriate name
```suggestion
    def open_analytic_lines(self):
```

mass_mailing_switzerland/models/utm.py
```diff
@@ -204,6 +204,31 @@ class UtmCampaign(models.Model):
...
+        return {
+            "name": _("Account Lines"),
```
ecino:
```suggestion
            "name": _("Analytic Lines"),
```

mass_mailing_switzerland/views/account_analytic_line_view.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<odoo>
+
+    <!-- Invoice Line Tree View -->
```
ecino:
```suggestion
    <!-- Analytic Line Tree View -->
```

ecino (changes requested):

ecino (approved):

PR #1427, opened 2021-11-18
CO-3829 : number of sponsorship on Mailchimp is wrong if many partner…

TheRaphael0000:
… linked

- add numspon in mass_mailing_contact which is the amount of sponsored
children for a email address (this new merge field should be added or changed on the production server)
- use the partner with the most children sponsored as the mailing recipient


TheRaphael0000:
I've just realized that I didnt fixed the problem with the 0 sponsorships

duvenhage (approved):

PR #1429, opened 2021-11-26
CS-636 : Events are not consistently linked with contacts/churches in Odoo

TheRaphael0000:
remove the feature that send email to attendee


ecino:
This is not needed.

PR #1430, opened 2021-11-29
CO-3830 : change of email trigger a merge of mass.mailing.contact

TheRaphael0000:
the remove condition was not correct and it should only remove the contract when these is no more partner in the contract

### DIFF
--- a/mass_mailing_switzerland/models/res_partner.py
+++ b/mass_mailing_switzerland/models/res_partner.py
@@ -105,12 +105,18 @@ class ResPartner(models.Model):
                 mailing_contact_vals["opt_out"] = vals["opt_out"]
             if "category_id" in vals:
                 mailing_contact_vals["tag_ids"] = vals["category_id"]
-            if "email" in vals and vals["email"] and not self.env.context.get(
-                    "import_from_mailchimp"):
+            # Whenever a partner with a mailing_contacts change email we update his mailchimp data
+            if "email" in vals and not self.env.context.get("import_from_mailchimp"):
+                # Remove the hold the partner from the mailing_contact
                 mailing_contacts.write({"partner_ids": [(3, self.id)]})
-                # In this case we don't need anymore the mailchimp contacts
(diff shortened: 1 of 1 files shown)


mass_mailing_switzerland/models/res_partner.py
```diff
@@ -105,12 +105,15 @@ def write(self, vals):
                 mailing_contact_vals["opt_out"] = vals["opt_out"]
             if "category_id" in vals:
                 mailing_contact_vals["tag_ids"] = vals["category_id"]
-            if "email" in vals and not vals["email"] and not self.env.context.get(
+            if "email" in vals and vals["email"] and not self.env.context.get(
```
ecino: I think we want to include the case were we remove the email address
```suggestion
            if "email" in vals and not self.env.context.get(
```

mass_mailing_switzerland/models/res_partner.py
```diff
@@ -105,12 +105,15 @@ def write(self, vals):
...
                     "import_from_mailchimp"):
+                mailing_contacts.write({"partner_ids": [(3, self.id)]})
                 # In this case we don't need anymore the mailchimp contacts
-                mailing_contacts.unlink()
+                if len(mailing_contacts.partner_ids) <= 0:
+                    mailing_contacts.unlink()
```
ecino, line 116: I think we miss a part to sync the new email to a mass mailing contact, no? Please add that.

ecino (changes requested):
I think the fix is wrong. The logic before the fix was to check if we remove the email from a partner, then we delete the mass mailing contacts associated.

Now you changed this to: if we change the email of a partner, the mailing contacts are removed. But I think that the new email address won't be exported to mailchimp!

mass_mailing_switzerland/models/res_partner.py
TheRaphael0000: I have add the code for this below (I dont understand why it won't ask a new review)

ecino (approved):

PR #1431, opened 2021-11-30
CO-3832 : add constrain on invoice validation

TheRaphael0000:
- Add a "raise UserError" when another account.invoice already is already open with the same transaction_id

### DIFF
--- a/sponsorship_switzerland/models/account_invoice.py
+++ b/sponsorship_switzerland/models/account_invoice.py
@@ -98,10 +98,10 @@ class AccountInvoice(models.Model):
         for invoice in self:
             condition = [
                 ("transaction_id", "=", invoice.transaction_id),
-                ("state", "=", "open"),
+                ("state", "in", ["open", "paid"]),
                 ("id", "!=", invoice.id),
             ]
-            if self.env[self._name].search(condition):
+            if self.env[self._name].search_count(condition):
                 error_msg = _("The Transaction ID `{}` is already used in another validated account invoice !")
                 raise UserError(error_msg.format(invoice.transaction_id))
         return super(AccountInvoice, self).action_invoice_open()
--- a/sponsorship_switzerland/models/account_invoice.py
+++ b/sponsorship_switzerland/models/account_invoice.py
@@ -93,3 +93,15 @@ class AccountInvoice(models.Model):
         """Reconcile given invoices with partner open payments.
         """
         return self._group_or_split_reconcile()
+
+    def action_invoice_open(self):
+        condition = [
+            ("transaction_id", "=", self.transaction_id),
+            ("state", "=", "open"),
+        ]
+        same_transactions_invoices = self.env["account.invoice"].search(condition)
+        others_same_transactions_invoices = same_transactions_invoices.filtered(lambda x: x.id != self.id)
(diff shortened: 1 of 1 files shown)


sponsorship_switzerland/models/account_invoice.py
```diff
@@ -93,3 +93,15 @@ def group_or_split_reconcile(self):
         """Reconcile given invoices with partner open payments.
         """
         return self._group_or_split_reconcile()
+
+    def action_invoice_open(self):
+        condition = [
+            ("transaction_id", "=", self.transaction_id),
```
ecino: Beware that in any method, `self` can be a recordset of multiple records (multiple invoices here). Accessing `self.transaction_id` may fail in such cases, so the best is to first loop over the recordset and test each invoice:
```
for invoice in self:
    invoice.transaction_id
```

sponsorship_switzerland/models/account_invoice.py
```diff
@@ -93,3 +93,15 @@ def group_or_split_reconcile(self):
...
+            ("state", "=", "open"),
+        ]
+        same_transactions_invoices = self.env["account.invoice"].search(condition)
+        others_same_transactions_invoices = same_transactions_invoices.filtered(lambda x: x.id != self.id)
```
ecino: Instead of filtering your results, you can add the condition directly on your search:
```suggestion
        condition = [
            ("transaction_id", "=", invoice.transaction_id),
            ("state", "=", "open"),
            ("id", "!=", invoice.id),
        ]
```

sponsorship_switzerland/models/account_invoice.py
```diff
@@ -93,3 +93,15 @@ def group_or_split_reconcile(self):
...
+        if len(others_same_transactions_invoices) > 0:
```
ecino: This works as well and is more clear:
```suggestion
        if others_same_transactions_invoices:
```

ecino (changes requested):
Small improvements can be done, otherwise the logic looks fine.

sponsorship_switzerland/models/account_invoice.py
```diff
@@ -93,3 +93,15 @@ def group_or_split_reconcile(self):
         """Reconcile given invoices with partner open payments.
         """
         return self._group_or_split_reconcile()
+
+    def action_invoice_open(self):
+        for invoice in self:
+            condition = [
+                ("transaction_id", "=", invoice.transaction_id),
+                ("state", "=", "open"),
```
ecino: Maybe we want to include paid invoices here.
```suggestion
                ("state", "in", ["open", "paid"]),
```

sponsorship_switzerland/models/account_invoice.py
```diff
@@ -93,3 +93,15 @@ def group_or_split_reconcile(self):
...
+                ("id", "!=", invoice.id),
+            ]
+            if self.env[self._name].search(condition):
```
ecino: This will be slighlty more performant and work the same (doing a count(*) query instead of fetching all data of records).
```suggestion
            if self.env[self._name].search_count(condition):
```

ecino (changes requested):

ecino (approved):

PR #1434, opened 2021-12-03
CO-3835 : Make "Write a letter" page available without login

TheRaphael0000:
Implement a magic route that create a user from a partner UUID and log it to my compassion

The magic route only work once (one time login) and only for partners without a My Compassion account.

If the user used this route and want to login again he must:
- Have access his mail and reset his My Compassion password with the reset password form
- Otherwise he must contact us to change his login from "magic_login_XXXX" to his actual mail address

Partner with a My Compassion account (even the one that used the route previously):
- Will be redirected to the login page

It would probably be possible to create a route that even work with partners that already have My Compassion account but it would require me more time to investigate on the authentications mechanism in Odoo.
Though I wouldn't recommend to create this route since it would probably create a security breach.

### DIFF
--- a/website_compassion/controllers/my_account.py
+++ b/website_compassion/controllers/my_account.py
@@ -234,7 +234,7 @@ class MyAccountController(PaymentFormController):
         # reset password
         crypt_context = CryptContext(schemes=["pbkdf2_sha512", "plaintext"], deprecated=["plaintext"])
         password_encrypted = crypt_context.encrypt(password)
-        request.env.cr.execute(f"UPDATE res_users SET password='{password_encrypted}' WHERE login='{login}';")
+        request.env.cr.execute("UPDATE res_users SET password=%s WHERE login=%s;", [password_encrypted, login])
         request.env.cr.commit()
 
         # authenticate
--- a/website_compassion/controllers/my_account.py
+++ b/website_compassion/controllers/my_account.py
@@ -17,6 +17,8 @@ from zipfile import ZipFile
 from urllib.request import urlretrieve, urlopen
 from math import ceil
 from dateutil.relativedelta import relativedelta
+import secrets
+from passlib.context import CryptContext
 
 from werkzeug.datastructures import Headers
 from werkzeug.wrappers import Response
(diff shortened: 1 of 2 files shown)


website_compassion/controllers/my_account.py
```diff
@@ -198,6 +200,49 @@ def _download_image(type, child_id=None, obj_id=None):
...
+        _, login, _ = res_users.signup(values=values, token=partner.signup_token)
+
+        # create a random password
+        password = secrets.token_urlsafe(16)
+
+        # reset password
+        crypt_context = CryptContext(schemes=["pbkdf2_sha512", "plaintext"], deprecated=["plaintext"])
+        password_encrypted = crypt_context.encrypt(password)
+        request.env.cr.execute(f"UPDATE res_users SET password='{password_encrypted}' WHERE login='{login}';")
+        request.env.cr.commit()
```
TheRaphael0000:
```suggestion
        user = res_users.search([["login", "=", login]], limit=1)
        user.sudo().write({"password": password_encrypted})
```
Notice that this code does't work and I don't know why...

TheRaphael0000: I changed the "execute" line to avoid SQL injections in the login

website_compassion/controllers/my_account.py
```diff
@@ -198,6 +200,72 @@ def _download_image(type, child_id=None, obj_id=None):
...
+    @route("/my/login/<partner_uuid>/<redirect_page>", type="http", auth="public", website=True)
+    def magic_login(self, partner_uuid=None, redirect_page=None):
+        if not partner_uuid:
+            return None
+
+        res_partner = request.env["res.partner"].sudo()
+        res_users = request.env["res.users"].sudo()
+
+        partner = res_partner.search([["uuid", "=", partner_uuid]], limit=1)
+        partner = partner.sudo()
```
ecino, line 212: Since you already did sudo at line 211, I think this one is not needed.
```suggestion
```

ecino (approved):

PR #1437, opened 2021-12-10
CO-3830: change of email trigger a merge of mass.mailing.contact

TheRaphael0000:
update the tags of the mass.mailing.contact with a computed field
when updating the res.partner email delete all mass.mailing.contact and recreate them


ecino:
I'm sorry but I think we will introduce more problems this way. I need to discuss this issue with David and rethink the process. For the moment I think it's too risky to delete contacts and recreate them, while bypassing mailchimp errors when we delete them. We will certainly be out of sync like that.

PR #1444, opened 2022-01-06
CS-656 Cannot see campaign results on mass mailing

TheRaphael0000:
- Add the get arguments from the mycompassion/letters to the XHR request to handle UTMs for letters
- Fix problem with the rights when using mailchimp id as the utm_campaign
- Simplify code flow

### DIFF
--- a/website_compassion/static/src/js/write_a_letter.js
+++ b/website_compassion/static/src/js/write_a_letter.js
@@ -354,6 +354,12 @@ async function sendLetter() {
         "Need": "${$('#child_id').text()}"
     }`);
 
+    let get_params = new URLSearchParams(window.location.search)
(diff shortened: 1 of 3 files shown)


website_compassion/static/src/js/write_a_letter.js
```diff
@@ -354,6 +354,12 @@ async function sendLetter() {
         "Need": "${$('#child_id').text()}"
     }`);
 
+    let get_params = new URLSearchParams(window.location.search)
+    get_params.forEach(function(v, k, _) {
+        json_data[k] = v;
+    })
+    console.log(json_data);
```
ecino: We don't need this in production.

ecino (approved):

PR #1449, opened 2022-01-12
CO-3853 update address from myCompassion failed and if success does n…

TheRaphael0000:
…ot change canton if needed

Geosptial add field called GeoPoint which can be directly affecty like this, this fix the problem


TheRaphael0000:
I had to use the new API key, that we generated with David, see on Jira

ecino (approved):

PR #1451, opened 2022-01-19
CO-3867 Simplify letter import

TheRaphael0000:
Remove old barcodes detection methods and implement new ones. Also remove the template detection and language detection.
See https://github.com/CompassionCH/compassion-modules/pull/1607


sbc_switzerland/models/import_letters_history.py
```diff
@@ -13,27 +13,20 @@
 """
 import base64
 import logging
-import subprocess
-import zipfile
-from io import BytesIO
-from tempfile import NamedTemporaryFile
+import traceback
+from pathlib import Path
```
ecino, line 17: Is that a new dependency added? If yes it should be added in the `requirements.txt` file and in module manifest under `external_dependencies` section.

TheRaphael0000: No both are from the python standard library:
https://docs.python.org/3/library/pathlib.html
https://docs.python.org/3/library/traceback.html

ecino (approved):

PR #1453, opened 2022-01-20
CO-3869 Some people don't receive the zoom link

TheRaphael0000:
I've refactor the code such that we don't have the time delta at two places. To do so I've added a computed Datetime based on the zoom date.

### DIFF
--- a/partner_communication_switzerland/models/res_partner_zoom_attendee.py
+++ b/partner_communication_switzerland/models/res_partner_zoom_attendee.py
@@ -93,6 +93,7 @@ class ZoomAttendee(models.Model):
                 user_id=user_id)
 
     def send_communication(self, config_name):
+        self.ensure_one()
         config_id = self.env.ref(config_name.value).id
         partner_id = self.partner_id.id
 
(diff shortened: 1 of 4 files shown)


TheRaphael0000:
To test and use my changes, please update the zoom automated action as following :

image

partner_communication_switzerland/models/res_partner_zoom_attendee.py
```diff
@@ -85,23 +92,33 @@ def notify_user(self):
...
+        else:
+            object_id = None
+
+        return self.env["partner.communication.job"].create({
+            "config_id": config_id,
+            "partner_id": partner_id,
+            "object_ids": object_id,
+        })
+
+    @api.one
```
ecino: api.one is deprectated and should not be used. but if you are sure you always call this method with a single record you can start the method with `self.ensure_one`

partner_communication_switzerland/models/res_partner_zoom_attendee.py
```diff
@@ -85,23 +92,33 @@ def notify_user(self):
...
-                config = short_notice_config
-                object_id = attendee.zoom_session_id.id
-            self.env["partner.communication.job"].create({
-                "partner_id": attendee.partner_id.id,
-                "config_id": config.id,
-                "object_ids": object_id
-            })
-        return True
+    def send_communication(self, config_name):
+        config_id = self.env.ref(config_name.value).id
```
ecino, line 97: Let's make it clear that we should only call this method on a single record.
```suggestion
        self.ensure_one()
        config_id = self.env.ref(config_name.value).id
```

ecino (changes requested):

ecino (approved):

PR #1455, opened 2022-01-24
CO-3877 create report for 1 QR Slip

TheRaphael0000:
add a paper size for single QR slip, adapt the QR slip template to allow to remove the scissors

### DIFF
--- a/report_compassion/report/paperformats.xml
+++ b/report_compassion/report/paperformats.xml
@@ -63,4 +63,20 @@
         <field name="header_spacing">3</field>
         <field name="dpi">96</field>
     </record>
+
+    <record id="paperformat_single_bvr_sponsorship" model="report.paperformat">
+        <field name="name">Single BVR Sponsorship</field>
+        <field name="default" eval="True"/>
+        <field name="format">custom</field>
+        <field name="orientation">Landscape</field>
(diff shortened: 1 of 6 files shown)


report_compassion/report/paperformats.xml
```diff
@@ -63,4 +63,20 @@
...
+        <field name="orientation">Landscape</field>
+        <field name="page_height">205</field>
+        <field name="page_width">110</field>
+        <field name="margin_top">0</field>
+        <field name="margin_bottom">0</field>
+        <field name="margin_left">0</field>
+        <field name="margin_right">0</field>
+        <field name="header_line" eval="False"/>
+        <field name="header_spacing">3</field>
+        <field name="dpi">46</field>
```
ecino, line 80: Normally we usde dpi 96 in order to get correct positioning in CSS with wkhtmltopdf.
```suggestion
        <field name="dpi">96</field>
```

ecino (changes requested):

report_compassion/report/paperformats.xml
TheRaphael0000: If we use the 96 dpi, the bill is not full page

ecino (approved):

PR #1456, opened 2022-01-25
CO-3873 Next invoice dates are set at the date of creation instead of…

TheRaphael0000:
… the 1st of the month for wordpress sponsorships

- Remove the next_invoice_date (which was set to today) when creating the sponsorship from Wordpress (thus use the default one which is correct)
**NOT TESTED**

- Create a migration which fix this problem in the current db

### DIFF
--- a/wordpress_connector/migrations/12.0.1.0.1/post-migration.py
+++ b/wordpress_connector/migrations/12.0.1.0.1/post-migration.py
@@ -1,0 +1,23 @@
+from openupgradelib import openupgrade
+from dateutil.relativedelta import relativedelta
+
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    if not version:
+        return
+
+    # Associate already created toilets fund to new xml record
+    sponsorships = env["recurring.contract"].search(
+        [("create_uid.name", "=", "WordPress Imports")]
+    )
+
+    for s in sponsorships:
+        d = s.next_invoice_date
+        if not d:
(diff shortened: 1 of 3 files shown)


wordpress_connector/migrations/12.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,23 @@
...
+
+@openupgrade.migrate(use_env=True)
+def migrate(env, version):
+    if not version:
+        return
+
+    # Associate already created toilets fund to new xml record
+    sponsorships = env["recurring.contract"].search(
+        [("create_uid.name", "=", "WordPress Imports")]
+    )
```
ecino:
```suggestion
    sponsorships = env["recurring.contract"].search([
        ("create_uid.name", "=", "WordPress Imports"),
        ("next_invoice_date", "!=", False),
        ("state", "not in", ["cancelled", "terminated"])
    ])
```

wordpress_connector/migrations/12.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,23 @@
...
+
+    for s in sponsorships:
+        d = s.next_invoice_date
+        if not d:
+            continue
```
ecino:
```suggestion
```

ecino (changes requested):

ecino (approved):

PR #1457, opened 2022-01-25
CO-3864 remove all suspension proccess for the invoices

TheRaphael0000:
remove inheritance that is not useful anymore


ecino:
This one is needed, it's to remove suspended children from the website, avoiding them to be sponsored during a suspension.

PR #1467, opened 2022-02-01
CO-3884 Event planned emails are not sent on the scheduled dates

TheRaphael0000:
Refactor the program flow to send email in a single function
Instead of commiting the postgres transaction at each email use savepoints (which only locally rollback data)
This ensure that odoo savepoints does not raise an Exceptions and rollback the whole transaction.

When one communication inside a schedule can't be send : we will try to send it each `Event: Mail Scheduler` call
We've added warning logs so those failing communication can be investigated without spamming the users

See also : https://github.com/CompassionCH/compassion-modules/pull/1613

### DIFF
--- a/website_event_compassion/models/event_email.py
+++ b/website_event_compassion/models/event_email.py
@@ -130,20 +130,17 @@ class EventMailRegistration(models.Model):
         for email in self:
             if email.mail_sent:
                 continue
-            try:
-                # if the communication fail by either an uncaught exception or is in the state failure
-                # skip this mail and dont mark it as sent
-                communication = self.env["partner.communication.job"].create(
-                    {
-                        "partner_id": email.registration_id.partner_id.id,
-                        "object_ids": email.registration_id.ids,
-                        "config_id": email.scheduler_id.communication_id.id,
-                    }
-                )
-                if communication is None or communication.state in ["failure"]:
-                    raise Exception
-                email.write({"mail_sent": True})
-            except Exception:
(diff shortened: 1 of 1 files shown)


website_event_compassion/models/event_email.py
```diff
@@ -136,15 +123,27 @@ def _compute_scheduled_date(self):
...
                     {
-                        "partner_id": registration.partner_id.id,
-                        "object_ids": registration.ids,
+                        "partner_id": email.registration_id.partner_id.id,
+                        "object_ids": email.registration_id.ids,
                         "config_id": email.scheduler_id.communication_id.id,
                     }
                 )
+                if communication is None or communication.state in ["failure"]:
+                    raise Exception
```
ecino: Why not putting the sent_to_everyone here instead of raising an exception catched just after?
```suggestion
                    sent_to_everyone = False
                    continue
```

ecino (changes requested):

website_event_compassion/models/event_email.py
TheRaphael0000: This cover also the case where the create communication raise an exception

ecino: I think the create method in communications already handle exceptions. I verified and when generating the text it catches them and puts the communication in failure state without raising the error.

I just saw one missing try/except when the number of pages is computed in method `def count_pdf_page`. One can be added there and here in the code I think we can remove this try/except clause.

ecino (approved):

PR #1473, opened 2022-02-07
CO-3868 fix dependencies for res.lang.compassion and langdetect

TheRaphael0000:


ecino (changes requested):
Please fix all the detect_lang calls :
image

ecino (approved):

PR #1474, opened 2022-02-08
CO-3896 notes should be displayed in the reconciliation view.

TheRaphael0000:
add a column in the reconciliation if the note exist


TheRaphael0000:
image

ecino (approved):

PR #1476, opened 2022-02-08
CO-3893 Add support language detection support for Letter

TheRaphael0000:
minor fix: ensure that the number of file computation is only for done in certain status

### DIFF
--- a/sbc_switzerland/models/import_letters_history.py
+++ b/sbc_switzerland/models/import_letters_history.py
@@ -97,12 +97,15 @@ class ImportLettersHistory(models.Model):
 
     @api.onchange("data", "import_folder_path")
     def _compute_nber_letters(self):
-        for letter in self.filtered(lambda l: l.state in ["draft", "pending", "open", "ready"]):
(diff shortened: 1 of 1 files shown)


sbc_switzerland/models/import_letters_history.py
```diff
@@ -97,7 +97,7 @@ class ImportLettersHistory(models.Model):
 
     @api.onchange("data", "import_folder_path")
     def _compute_nber_letters(self):
-        for letter in self:
+        for letter in self.filtered(lambda l: l.state in ["draft", "pending", "open", "ready"]):
```
ecino:
```
        to_check = self.filtered(lambda l: l.state != "done")
        others = self - to_check
        for letter in to_check:
            ....
            except (FileNotFoundError, TypeError):
                others += letter
        super(ImportLettersHistory, others)._compute_nber_letters()
```

ecino (changes requested):
This way we can compute files using the traditional method as well.

ecino (approved):

PR #1479, opened 2022-02-09
CS-717 tax receipt for company/contact is wrong

TheRaphael0000:
add invoices from the company, when the tax receipt is computed from an employee

### DIFF
--- a/report_compassion/models/res_partner.py
+++ b/report_compassion/models/res_partner.py
@@ -40,9 +40,16 @@ class ResPartner(models.Model):
                 ("last_payment", "<=", end_date),
                 ("state", "=", "paid"),
                 ("product_id.requires_thankyou", "=", True),
+                # invoice from either the partner, the company or the employee
+                # to obtain the same results when tax receipt is computed from companies or employees
                 "|",
+                # invoice from the partner (when self is either an company or an employee)
                 ("partner_id", "=", self.id),
+                "|",
+                # invoice from the company (when self is an employee)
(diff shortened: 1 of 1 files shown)


ecino (approved):
@davidwul Can you review as well?

ecino (dismissed):

report_compassion/models/res_partner.py
```diff
@@ -40,9 +40,16 @@ def get_receipt(self, year):
...
                 ("state", "=", "paid"),
                 ("product_id.requires_thankyou", "=", True),
+                # invoice from either the partner, the company or the employee
+                # to obtain the same results when tax receipt is computed from companies or employees
                 "|",
+                # invoice from the partner (when self is either an company or an employee)
                 ("partner_id", "=", self.id),
+                "|",
+                # invoice from the company (when self is an employee)
                 ("partner_id.parent_id", "=", self.id),
```
ecino, line 50: @davidwul Should we remove this, in order for tax receipt to show 0.- if the invoice is at the company level?

report_compassion/models/res_partner.py
```diff
@@ -40,9 +40,16 @@ def get_receipt(self, year):
...
+                # invoice from the employees (when self is a company)
+                ("partner_id.child_ids", "=", self.id),
```
ecino, line 52: @davidwul It's strange that this had to be added, because it's the opposite of the task description in which you said "tax receipt for companies are correct".

PR #1483, opened 2022-02-10
CS-691 Wrong link

TheRaphael0000:
- create a function to generate the gift link depending on the child
- use it on the my children page

### DIFF
--- a/website_compassion/models/res_user.py
+++ b/website_compassion/models/res_user.py
@@ -7,9 +7,8 @@
 #
 ##############################################################################
 import logging
-import urllib.parse
 
-from odoo import models, fields, _
+from odoo import models, fields
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


website_compassion/models/res_user.py
```diff
@@ -31,3 +32,27 @@ def reset_password(self, login):
...
+            "pcity": partner.city,
+            "pzip": partner.zip,
+            "pcountry": partner.country_id.name,
+            "sponsor_ref": partner.ref,
+            "child_ref": child.local_id,
+        }
+        filter_data = {key: val for key, val in data.items() if val}
+        params = urllib.parse.urlencode(filter_data)
+        url = f"{_('https://compassion.ch/de/geschenkformular/')}?{params}"
+        return url
```
ecino: This is not needed. You can use the compute field already existing `partner.wordpress_form_data`

ecino (changes requested):

website_compassion/models/res_user.py
TheRaphael0000: I don't think I can. We want the link to be available for any child, with the suggested function it doesn't work if the sponsor have multiple children. See below

https://github.com/CompassionCH/compassion-switzerland/blob/4a47c9a99524d79072ee4d8a234e4c98ddb25adf/mass_mailing_switzerland/models/res_partner.py#L64

TheRaphael0000: I think it would be best to call, my new function with this condition to avoid code duplication

`child.local_id if len(child) == 1 else False
`

ecino: You can put the child to use in the context variable `mailchimp_child` (I know, the name is not great, we could rename this variable if needed). It's read here : https://github.com/CompassionCH/compassion-switzerland/blob/4a47c9a99524d79072ee4d8a234e4c98ddb25adf/mass_mailing_switzerland/models/res_partner.py#L50

ecino (approved):

PR #1484, opened 2022-02-11
CO-3904 update myAccount throws an error

TheRaphael0000:
- add first res_user login to the partner view
- remove the login update on partner update
- add the sudo to the self [res.partner] since some users couldn't change their name (due to access rights on calendar.attendee)
- show the login on MyCompassion (but no edit since it introduce too many bugs)
- refactor the my/information route
- divers refactor (mostly auto format)

### DIFF
--- a/partner_compassion/models/partner_compassion.py
+++ b/partner_compassion/models/partner_compassion.py
@@ -340,8 +340,6 @@ class ResPartner(models.Model):
             if len(users) > 0:
                 user = users[0]
                 user.login = partner.user_login
-            else:
-                pass
 
     @api.multi
     @api.depends("segments_affinity_ids", "segments_affinity_ids.affinity")
--- a/partner_compassion/models/partner_compassion.py
+++ b/partner_compassion/models/partner_compassion.py
@@ -388,7 +388,6 @@ class ResPartner(models.Model):
 
     @api.multi
     def write(self, vals):
-        self = self.sudo()
         if vals.get("criminal_record"):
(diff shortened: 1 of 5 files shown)


partner_compassion/models/partner_compassion.py
```diff
@@ -317,6 +324,25 @@ def _compute_thankyou_letter(self):
...
+
+    @api.multi
+    def _set_user_login(self):
+        for partner in self:
+            users = partner.user_ids
+            if len(users) > 0:
+                user = users[0]
+                user.login = partner.user_login
+            else:
+                pass
```
ecino: This can be removed.
```suggestion
```

partner_compassion/models/partner_compassion.py
```diff
@@ -364,16 +390,7 @@ def create(self, vals):
...
-        if email:
-            vals["email"] = email.strip()
-            # Push email to non-internal users
-            user_ids = self.mapped("user_ids").filtered("share").ids
-            self.env.cr.execute("""
-                UPDATE res_users
-                SET login=%s
-                WHERE id=ANY(%s)
-            """, [vals["email"], user_ids])
+        self = self.sudo()
```
ecino: I don't think using sudo in the write method is a good idea, it could lead to many security flaws. I would rather add the missing security rules if some are needed.
```suggestion
```

ecino (changes requested):

ecino (approved):

PR #1311, opened 2021-04-07
CO-3594: -Added a new subsection "My sponsorship" in "My Financials".

coup1234:

PR #1487, opened 2022-02-15
CO-3594 Changer la vue des options de paiement

TheRaphael0000:
Rebase devel from the previous PR, fix conflicts
Complement the code from a previous PR to allow to upgrade standard sponsorship and write&pray


ecino:
Superseeds #1311

PR #1488, opened 2022-02-15
CO-3907 Corrections of the new postfinance gateway

TheRaphael0000:
- fix invoice duplicate problem by removing the create call on form creation (keep the one on update)
- add an autocancel date for the newly created invoices


website_event_compassion/models/account_invoice.py
```diff
@@ -53,3 +61,9 @@ def action_invoice_paid(self):
         task = self.env.ref("website_event_compassion.task_full_payment")
         registrations.write({"completed_task_ids": [(4, task.id)]})
         return res
+
+    @api.multi
+    def auto_cancel(self):
+        invoices = self.filtered(lambda i: i.state in ["open"])
```
ecino: Maybe we should as well consider cancelling draft invoices?

```suggestion
        invoices = self.filtered(lambda i: i.state in ["draft", "open"])
```

ecino (changes requested):
Why not moving this code into the module `cms_form_compassion` so that we can use the functionality in every form donation we would like?

ecino (approved):

PR #1496, opened 2022-02-23
CO-3795 Cannot delete a bounced mailchimp contact

TheRaphael0000:
catch exception and log it in error instead of stopping the program

### DIFF
--- a/mass_mailing_switzerland/models/mass_mailing_contact.py
+++ b/mass_mailing_switzerland/models/mass_mailing_contact.py
@@ -226,8 +226,8 @@ class MassMailingContact(models.Model):
     def unlink(self):
         try:
             self.action_archive_from_mailchimp()
-        except Exception as e:
-            logger.error(traceback.format_exc())
+        except:
+            logger.error("Error archiving mailchimp contact", exc_info=True)
         return super().unlink()
 
     ##########################################################################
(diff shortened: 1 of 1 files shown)


ecino (dismissed):

mass_mailing_switzerland/models/mass_mailing_contact.py
```diff
@@ -220,7 +224,10 @@ def write(self, values):
 
     @api.multi
     def unlink(self):
-        self.action_archive_from_mailchimp()
+        try:
+            self.action_archive_from_mailchimp()
+        except Exception as e:
+            logger.error(traceback.format_exc())
```
ecino: This way we don't need the `traceback` library.
```suggestion
        except:
            logger.error("Error archiving mailchimp contact", exc_info=True)
```

mass_mailing_switzerland/models/mass_mailing_contact.py
```diff
@@ -9,12 +9,16 @@
 ##############################################################################
 from ast import literal_eval
 from datetime import date
+import logging
+import traceback
```
ecino:
```suggestion
```

ecino (changes requested):

ecino (approved):

PR #1505, opened 2022-03-21
CO-3922: Convert amount raised to company currency

ovesco:
Using invoice_line's price_subtotal_signed which includes the amount already converted to the company's currency.

- On the Muskathlon page, display converted raised amount per participant by using price_subtotal_signed in raw SQL query
- in the event_registration, also use price_subtotal_signed
- update hardcoded CHF field in various email templates
- add indexes on user_id and event_id in muskathlon_report, it should speed up queries a bit

Please note that I'm not sure if the changes cover everything


muskathlon/data/mail_template.xml
```diff
@@ -42,7 +42,7 @@
                     ${partner.salutation},
                     <br/>
                     <br/>
-                    Compassion Switzerland is glad to inform you that you just received a donation of CHF ${donations[0]}.-
+                    Compassion Switzerland is glad to inform you that you just received a donation of ${donations[0].currency_id.symbol} ${donations[0]}.-
```
ecino:
```suggestion
                    Compassion Switzerland is glad to inform you that you just received a donation of ${object.user_id.company_id.currency_id.symbol} ${donations[0]}.-
```

ecino (changes requested):

muskathlon/models/event_registration.py
```diff
@@ -67,13 +67,14 @@ def _compute_step2_tasks(self):
         for reg in self:
             reg.passport_uploaded = muskathlon_passport in reg.completed_task_ids
 
-    def _compute_amount_raised(self):
+    def _compute_amount_raised(self, force_compute=False):
```
ecino: I think `force_compute` is no longer used, you could remove it.

ovesco: Oh yes indeed

ecino (approved):

PR #1509, opened 2022-04-05
Write and pray journey

ecino:
When we go live:

- Change urls of communications
- Add date filter in automations to avoid sending communications to all previous WP sponsorships


ecino:
When we go live:

- Change urls of communications
- Add date filter in automations to avoid sending communications to all previous WP sponsorships

PR #1515, opened 2022-04-26
Co 3948 Warn upon create_sponsorship_job that fails

ovesco:
Not sure if this is the correct approach, did not find any reference to opsgenie in code but rather sending emails so reproduced it here.

This commit adds code to catch any error occuring in the create_sponsorship_job method of the wordpress connector and sends an email notifying of the problem.

### DIFF
--- a/wordpress_connector/models/contracts.py
+++ b/wordpress_connector/models/contracts.py
@@ -251,75 +251,89 @@ class Contracts(models.Model):
         """
         child = self.env["compassion.child"].browse(values["child_id"])
         child.remove_from_wordpress()
-        sponsorship = self.env["recurring.contract"].create(values)
-        list_keys = [
-            "salutation",
-            "first_name",
-            "last_name",
-            "birthday",
-            "street",
-            "zipcode",
-            "city",
-            "land",
-            "email",
-            "phone",
-            "lang",
-            "language",
-            "kirchgemeinde",
-            "Beruf",
-            "zahlungsweise",
-            "consumer_source",
-            "consumer_source_text",
-            "patenschaftplus",
-            "mithelfen",
-            "childID",
-            "Child reference",
-        ]
-        web_info = ""
-        for key in list_keys:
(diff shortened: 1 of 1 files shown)


wordpress_connector/models/contracts.py
```diff
@@ -251,75 +251,89 @@ def create_sponsorship_job(self, values, form_data):
...
 
-        sponsorship.correspondent_id.set_privacy_statement(origin="new_sponsorship")
-        return sponsorship
+            sponsorship.correspondent_id.set_privacy_statement(origin="new_sponsorship")
+            return sponsorship
+        except BaseException as err:
+            """Log the error and send a mail stating that it failed running"""
+            _logger.error("Wordpress create sponsorship job failed", exc_info=True)
+            child.activity_schedule(
+                'mail.mail_activity_data_todo',
```
ecino:
```suggestion
                'mail_activity_data_warning',
```

wordpress_connector/models/contracts.py
```diff
@@ -251,75 +251,89 @@ def create_sponsorship_job(self, values, form_data):
...
+                summary="[URGENT] Wordpress create sponsorship job failed",
+                note="Please verify this new sponsorship made from the website with "
+                     f"following information: {form_data}, "
+                     f"and given following values: {values}, "
+                     f"with following error: {err}",
```
ecino: Does this work or should we cast to string? Not sure how f-strings work.
```suggestion
                     f"following information: {str(form_data)}, "
                     f"and given following values: {str(values)}, "
                     f"with following error: {str(err)}",
```

ecino (changes requested):

wordpress_connector/models/contracts.py
ovesco: According to documentation (and simple tests), f-strings work nicely with dicts, but we can cast for safety

ecino (approved):

PR #1522, opened 2022-05-09
Co 3965     Schedule activity when a sponsorship is validated and the partner has "don't inform" as a communication setting

ovesco:
This PR schedules an activity when a sponsorship's state becomes "waiting" and the sponsor's communication settings are set to "don't inform". 
- The activity is set with a max due time of one week
- Assigned to the SDS person responsible for the language of the sponsor
- Migration to add activity to sponsors with "dont inform" (Please check migration number!)
- Fix a potential bug in contract where payment_mode might be False

Please note that the migration is special in this case, as we have to add an activity to a set of sponsors only once, when the update is deployed. As such it is not really a migration nor should it be registered in a module lifecycle hook...
It would have been possible to use something like this: https://www.odoo.com/fr_FR/forum/aide-1/can-we-invoke-a-function-automatically-while-installing-a-module-165800 but would have implied adding a method in the model which would have run only once before becoming dead code. As such the code is put in a migration.


partner_communication_switzerland/migrations/12.0.1.1.4/post-migration.py
```diff
@@ -0,0 +1,20 @@
...
+@openupgrade.migrate(use_env=True)
+def migrate(env, installed_version):
+    if not installed_version:
+        return
+
+    # Add activities "for dont inform" sponsors
+    sponsorships = env["res.partner"]\
+        .search([('global_communication_delivery_preference', '=', 'none')])\
+        .sponsorship_ids\
+        .filtered(lambda it: it.state in ["waiting", 'active'])
```
ecino: I would rather only use a search query. We also want to exclude a special partner that has more than 500 sponsorships.
```suggestion
    sponsorships = env["recurring.contract"].search([
        ('correspondent_id.global_communication_delivery_preference', '=', 'none'),
        ('state', 'in', ['waiting','active']),
        ('correspondent_id.ref', '!=', '1502623')])
```

partner_communication_switzerland/models/contracts.py
```diff
@@ -483,6 +483,25 @@ def contract_waiting_mandate(self):
...
+        config_settings = self.env["res.config.settings"].sudo()
+        sds_partner_id = config_settings.get_sponsorship_de_id()
+        if self.partner_id.lang == 'fr_CH':
+            sds_partner_id = config_settings.get_sponsorship_fr_id()
+        if self.partner_id.lang == 'it_IT':
+            sds_partner_id = config_settings.get_sponsorship_it_id()
+        sds_user = self.env['res.users'].sudo().search([('partner_id', '=', int(sds_partner_id))])
+
+        if sds_user.id:
+            self.partner_id.activity_schedule(
```
ecino: Let's use `correspondent` instead of payer.
```suggestion
        if self.correspondent_id.lang == 'fr_CH':
            sds_partner_id = config_settings.get_sponsorship_fr_id()
        if self.correspondent_id.lang == 'it_IT':
            sds_partner_id = config_settings.get_sponsorship_it_id()
        sds_user = self.env['res.users'].sudo().search([('partner_id', '=', int(sds_partner_id))])

        if sds_user.id:
            self.correspondent_id.activity_schedule(
```

partner_communication_switzerland/models/contracts.py
```diff
@@ -494,6 +513,12 @@ def contract_waiting(self):
                 and c.start_date < contract.start_date)
             contract.is_first_sponsorship = not old_sponsorships
 
+            # Notify SDS when partner has "don't inform" as comm setting
+            if contract.partner_id.global_communication_delivery_preference == "none":
```
```suggestion
            if contract.correspondent_id.global_communication_delivery_preference == "none":
```

ecino (changes requested):

partner_communication_switzerland/migrations/12.0.1.1.4/post-migration.py
ovesco: Alright, and damn for the special partner!

ecino (approved):

PR #1525, opened 2022-05-17
CS-736:  address duplication in Odoo

ovesco:
- refactor the duplicate checks from the create and onchance methods in a separate one to build initial domain (I want my A from SolarCloud!)
- build checks progressively in a way that it is easy to add new checks later on
- add a check to verify more thoroughly similar addresses

### DIFF
--- a/partner_compassion/models/partner_compassion.py
+++ b/partner_compassion/models/partner_compassion.py
@@ -369,25 +369,9 @@ class ResPartner(models.Model):
     ##########################################################################
     @api.model
     def create(self, vals):
-        """
-        Lookup for duplicate partners and notify.
-        """
-        email = vals.get("email")
-        if email:
-            vals["email"] = email.strip()
-        duplicate = self.search(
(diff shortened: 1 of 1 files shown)


partner_compassion/models/partner_compassion.py
```diff
@@ -517,47 +501,82 @@ def _check_qorder(self, word):
...
+                ("street", "ilike", vals.get('street'))
+            ])
+        ]
+
+        # This step builds a domain query based on the checks that
+        # passed, prepending the list with a "|" operator for all item
+        # once its size is > 1
+        search_filters = []
+        for check in checks:
+            if skip_props_check or check[0]:
```
ecino, line 550: Shouldn't be this `if not skip_props_check and check[0]:`, or is the variable name confusing? I'm not sure to understand the use of it, maybe a method comment would be helpful with parameters description.

ecino (approved):

PR #1527, opened 2022-05-24
Co 3971: reference textuel pour des contrat avec virement

ovesco:
This PR adds the possibility to create recurring contracts that are automatically reconcilied with recurring payments from a sponsor/donor.

To use it:
- Check in the bank statement line the "label", and copy the beginning of it, enough to be unique
- Create a new recurring contract for the donor
- Set the payment option to "wire transfer", this allows you to set any BVR reference, put the copied beginning of the bank statement label

Please note:
I am absolutely NOT sure about the sequence of the completion rule, I put 20 because it worked in my tests... Sometimes my completion rule was not checked because another more important one yielded a result before.

### DIFF
--- a/sponsorship_switzerland/models/completion_rules.py
+++ b/sponsorship_switzerland/models/completion_rules.py
@@ -152,7 +152,6 @@ class StatementCompletionRule(models.Model):
             return res
 
         recurring_contract_groups = self.env['recurring.contract.group'].search([
-            "&",
             ("bvr_reference", "!=", False),
             ("payment_mode_id", "=", wire_transfer_mode.id)
         ]).filtered(lambda it: name.startswith(it.bvr_reference))
--- a/sponsorship_switzerland/models/completion_rules.py
+++ b/sponsorship_switzerland/models/completion_rules.py
@@ -52,6 +52,11 @@ class StatementCompletionRule(models.Model):
             ),
             ("get_from_lsv_dd", "Compassion: Put LSV DD Credits in 1098"),
             (
+                "recurring_contract_wire_transfer_mapping",
+                "Compassion: From custom reference on wire transfer"
+                "(based on the CAMT label)",
+            ),
+            (
                 "get_sponsor_name",
                 "Compassion[POST]: From sponsor reference "
                 "(based the sponsor name in the description)",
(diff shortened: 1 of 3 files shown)


sponsorship_switzerland/models/completion_rules.py
```diff
@@ -131,6 +136,32 @@ def get_from_partner_ref(self, stmts_vals, st_line):
...
         return res
 
+    def recurring_contract_wire_transfer_mapping(self, stmts_val, st_line):
+        if "name" not in st_line:
+            return
+
+        name = st_line['name']
+        res = dict()
+
+        wire_transfer_mode = self.env['account.payment.mode'].search([
```
ecino: The name is translated, we want to find it even if the user importing is not using English.
```suggestion
        wire_transfer_mode = self.env['account.payment.mode'].with_context(lang="en_US").search([
```

sponsorship_switzerland/models/completion_rules.py
```diff
@@ -131,6 +136,32 @@ def get_from_partner_ref(self, stmts_vals, st_line):
...
+            ("name", "=", "wire transfer")
+        ])
+
+        if not wire_transfer_mode:
+            logger.warning("Unable to find wire transfer payment mode")
+            return res
+
+        recurring_contract_groups = self.env['recurring.contract.group'].search([
+            "&",
```
ecino: "and" is the default operator for any conditions and can be avoided.
```suggestion
```

sponsorship_switzerland/models/completion_rules.py
```diff
@@ -131,6 +136,32 @@ def get_from_partner_ref(self, stmts_vals, st_line):
...
+            ("bvr_reference", "!=", False),
+            ("payment_mode_id", "=", wire_transfer_mode.id)
+        ]).filtered(lambda it: name.startswith(it.bvr_reference))
+
+        if recurring_contract_groups:
+            res['partner_id'] = recurring_contract_groups.partner_id
```
ecino: Just in case we would have several results.
```suggestion
            res['partner_id'] = recurring_contract_groups[0].partner_id
```

sponsorship_switzerland/models/contract_group.py
```diff
@@ -189,6 +189,13 @@ def on_change_payment_mode(self):
     def on_change_bvr_ref(self):
         """ Test the validity of a reference number. """
         bvr_reference = self.bvr_reference
+
+        # Omit check if payment mode is wire transfer so that we can manually
+        # create recurring contracts based on recurrent payment a partner does
+        # reference: CO-3971
+        if self.payment_mode_id.name == 'wire transfer':
```
ecino:
```suggestion
        if self.payment_mode_id.with_context(lang="en_US").name == 'wire transfer':
```

ecino (changes requested):

sponsorship_switzerland/models/completion_rules.py
ovesco: Indeed

ovesco:
Thanks for reviewing!

ecino (approved):

PR #1533, opened 2022-07-28
Seb cp5

sebastiencompassion:


ecino:
You should rebase this branch onto devel so that it contains only your two commits regarding CP-5 and drop the two commits coming from the sbc-translation branch. (Using rebase interactive)

ecino:
This does not resolve the issue. We will postpone the issue for a future sprint.

PR #1535, opened 2022-09-28
cp-111 remove donors from onboarding when merged with sponsor

duvenhage:


partner_compassion/wizards/partner_merge_wizard.py
```diff
@@ -41,6 +41,14 @@ def action_merge(self):
                     "to send new labels to them."
                 )
             )
+        # check onboarding_new_donor_start_date for non-dst partner. If set,
+        # and dst partner is sponsor, clear the onboarding_new_donor_start_date.
+        if removing.onboarding_new_donor_start_date:
+            for obj in self.dst_partner_id.category_id:
+                if obj.display_name == 'Sponsor':
+                    removing.onboarding_new_donor_start_date = ''
+                    break
```
ecino: I propose this instead, using the record xml_id of the Sponsor category:
```suggestion
            sponsor_category = self.env.ref("partner_compassion.res_partner_category_sponsor")
            if sponsor_category in self.dst_partner_id.category_id:
                removing.onboarding_new_donor_start_date = False
```

ecino (changes requested):

PR #1536, opened 2022-11-21
FIX: correct pictures condition

NoeBerdoz:
The SDS reported a problem with the generation of the sponsorship certificates. [CP-236](https://compassion-ch.atlassian.net/browse/CP-236)

In many cases, there wasn't 2 pictures in the certificate even if the child was having many pictures. 

The XML condition was checking the exact date of the pictures, but in many cases, the first picture year of the child is greater then the contract start year.

### DIFF
--- a/report_compassion/report/ending_sponsorship_certificate.xml
+++ b/report_compassion/report/ending_sponsorship_certificate.xml
@@ -106,7 +106,7 @@
             <t t-foreach="docs" t-as="o">
                 <t t-call="report_compassion.style"/>
                 <t t-set="pictures" t-value="o.child_id.pictures_ids"/>
-                <t t-set="first_pic_available" t-value="len(pictures) &gt; 1 and pictures[-1].date.year &gt;= o.start_date.year"/>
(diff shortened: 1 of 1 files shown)


report_compassion/report/ending_sponsorship_certificate.xml
```diff
@@ -106,7 +106,7 @@
             <t t-foreach="docs" t-as="o">
                 <t t-call="report_compassion.style"/>
                 <t t-set="pictures" t-value="o.child_id.pictures_ids"/>
-                <t t-set="first_pic_available" t-value="len(pictures) &gt; 1 and pictures[-1].date.year == o.start_date.year"/>
+                <t t-set="first_pic_available" t-value="len(pictures) &gt; 1 and pictures[-1].date.year &gt;= o.start_date.year"/>
```
ecino: Should we not also test that the picture is older than the last one?
```suggestion
                <t t-set="first_pic_available" t-value="len(pictures) &gt; 1 and pictures[-1].date.year &gt;= o.start_date.year and pictures[-1].date.year &lt; pictures[0].date.year"/>
```

PR #1540, opened 2023-02-14
CP-143 let SDS manage if they send the intro letter to the sponsor

simgonzalez:
Allow SDS to manage if the intro letter is sent or not for a specific sponsorship.

### DIFF
--- a/partner_communication_switzerland/models/contracts.py
+++ b/partner_communication_switzerland/models/contracts.py
@@ -88,13 +88,6 @@ class RecurringContract(models.Model):
         else:
             self.send_introduction_letter = True
 
-    @api.multi
-    def _on_change_correspondant(self, correspondent_id):
-        # Don't send introduction letter when correspondent is changed
-        cancelled_sponsorships = super()._on_change_correspondant(correspondent_id)
-        cancelled_sponsorships.write({"send_introduction_letter": False})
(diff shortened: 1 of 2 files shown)


partner_communication_switzerland/models/contracts.py
```diff
@@ -88,13 +88,6 @@ def _do_not_send_letter_to_transfer(self):
         else:
             self.send_introduction_letter = True
 
-    @api.multi
-    def _on_change_correspondant(self, correspondent_id):
-        # Don't send introduction letter when correspondent is changed
-        cancelled_sponsorships = super()._on_change_correspondant(correspondent_id)
-        cancelled_sponsorships.write({"send_introduction_letter": False})
-        return cancelled_sponsorships
-
```
ecino, line 97: I would still keep the onchange method so that by default it disables the setting, while the user still has the final control on the value he wants to put and can still re-activate it if needed.

ecino (changes requested):

partner_communication_switzerland/models/contracts.py
ecino, line 97:
```
@api.onchange("correspondent_id")
def onchange_correspondent(self):
    self.send_introduction_letter = False
    return True
```

ecino (approved):

PR #1546, opened 2023-03-23
T0027 migrate child sync wp

simgonzalez:


simgonzalez:
This PR couldn't been tested without the migration of child switzerland.

PR #1560, opened 2023-05-23
T0190 cancel of old invoices when generating waiting reminder

simgonzalez:


ecino:
> invoices should be cancelled but on the first payment or when the sponsorship is cancelled

That part doesn't seem to have been implemented. Is it still needed or already working as it is in the code?

simgonzalez:
@ecino This part is something done in the contract_active (not sure about the exact name) method when the sponsorship state is updated.

ecino (approved):

PR #1565, opened 2023-07-17
14.0 sponsorship switzerland

ecino:
Module migration

Removed some features that were outdated for v14. It will be easier to start from scratch if they appear to be missing (to be determined during testing phase in stage that should be done by the accounting team). It's mostly about LSV/DD invoices generation.


sponsorship_switzerland/models/contract_group.py
```diff
@@ -0,0 +1,236 @@
...
+        count_groups = str(self.search_count([("partner_id", "=", partner.id)]))
+        result += "0" * (5 - len(count_groups)) + count_groups
+        # Type '0' = Sponsorship
+        result += "0"
+        result += "0" * 4
+
+        if is_lsv:
+            result = "004874969" + result[9:]
+        if len(result) == 26:
+            return mod10r(result)
```
NoeBerdoz, line 125: If you have the context of all these magic numbers, it would be good to have a few comments explanation,
As this file path doesn't exist anymore right? (/nas/it/devel/Code_ref_BVR.xls pointed in the function comment)

NoeBerdoz (approved):

sponsorship_switzerland/models/contract_group.py
ecino: Basically it's a few custom-made rules for generating a BVR reference. David made them I think, we should ask him if he still has the file. We can maybe include the file inside the module for reference. I will assign me a task for that.

PR #1567, opened 2023-07-18
14.0 report compassion

ecino:
Depends on #1565 .


NoeBerdoz (approved):

report_compassion/models/contract.py
```diff
@@ -0,0 +1,102 @@
...
+
+    _inherit = ["account.move", "translatable.model"]
+    _name = "account.move"
+
+    # Gender field is mandatory for translatable models
+    gender = fields.Char(compute="_compute_gender")
+
+    def _compute_gender(self):
+        for i in self:
+            i.gender = "M"
```
NoeBerdoz, line 27: I understand that it's currently the way to make it translatable, but it is ugly to do so in my opinion, isn't there a better way? I mean, it's weird to have a gender field for an AccountInvoice class

PR #1581, opened 2024-02-06
add needed field to handle salary in odoo

davidwul:
instead of mirgating payroll on l10n_ch


ecino (approved):

ecino:
Merged manually. :+1:

PR #1582, opened 2024-02-26
14.0 t0290

auclerc:


partner_compassion/models/partner_compassion.py
```diff
@@ -522,6 +529,59 @@ def action_view_partner_invoices(self):
         action["context"] = {"search_default_partner_id": self.id}
         return action
 
+    def check_phone_and_mobile(self, vals):
+        all_phone_destination_codes = [
```
ecino: Where does this list come from? Maybe a comment would be helpful here.

partner_compassion/models/partner_compassion.py
```diff
@@ -522,6 +529,59 @@ def action_view_partner_invoices(self):
...
+            52,
+            55,
+            56,
+            58,
+            61,
+            62,
+            71,
+        ]
+
+        all_mobile_destination_codes = [74, 75, 76, 77, 78, 79]
```
ecino: Where does it come from?

partner_compassion/models/partner_compassion.py
```diff
@@ -522,6 +529,59 @@ def action_view_partner_invoices(self):
...
+
+        phone = vals.get("phone")
+        mobile = vals.get("mobile")
+
+        if phone:
```
ecino: Maybe we should also check if the partner country is switzerland before doing the check. We may sometimes create partners from other countries that may have a different phone format. Or we can maybe validate the phone for any country with the phonenumbers library?

partner_compassion/models/partner_compassion.py
```diff
@@ -522,6 +529,59 @@ def action_view_partner_invoices(self):
...
+            parsed_phone = phonenumbers.parse(phone, "CH")
+            if not phonenumbers.is_valid_number(parsed_phone):
+                raise UserError(_("Phone number is not valid."))
+            phone_national_destination_code = int(str(parsed_phone.national_number)[:2])
+            if phone_national_destination_code in all_mobile_destination_codes:
+                vals["mobile"] = phone
+                if not mobile:
+                    vals["phone"] = False
+
+        if mobile:
```
ecino: Same remark about checking the country.

partner_compassion/models/partner_compassion.py
```diff
@@ -522,6 +529,59 @@ def action_view_partner_invoices(self):
...
+        all_mobile_destination_codes = [74, 75, 76, 77, 78, 79]
+
+        phone = vals.get("phone")
+        mobile = vals.get("mobile")
+
+        if phone:
+            parsed_phone = phonenumbers.parse(phone, "CH")
+            if not phonenumbers.is_valid_number(parsed_phone):
+                raise UserError(_("Phone number is not valid."))
+            phone_national_destination_code = int(str(parsed_phone.national_number)[:2])
```
ecino: I don't understand why you need such conversions. What is the format of `parsed_phone.national_number`?

partner_compassion/models/partner_compassion.py
```diff
@@ -522,6 +529,59 @@ def action_view_partner_invoices(self):
...
+            if phone_national_destination_code in all_mobile_destination_codes:
+                vals["mobile"] = phone
+                if not mobile:
```
ecino: Why is this `if` needed, shouldn't we remove the `phone` value in any case, if it's moved in the `mobile` field?

partner_compassion/models/partner_compassion.py
```diff
@@ -522,6 +529,59 @@ def action_view_partner_invoices(self):
...
+            parsed_mobile = phonenumbers.parse(mobile, "CH")
+            if not phonenumbers.is_valid_number(parsed_mobile):
+                raise UserError(_("Mobile number is not valid."))
+            mobile_national_destination_code = int(
+                str(parsed_mobile.national_number)[:2]
+            )
+            if mobile_national_destination_code in all_phone_destination_codes:
+                vals["phone"] = mobile
+                if not phone:
+                    vals["mobile"] = False
```
ecino: Maybe we should better check if a `phone` value is already set and not override it, and also check that we didn't move the `phone` field value and avoid deleting it afterward (you should introduce the bool var for that purpose).
```suggestion
                if not vals.get("phone"):
                    vals["phone"] = mobile
                    if not phone_moved_to_mobile:
                        vals["mobile"] = False
```

partner_compassion/models/partner_compassion.py
```diff
@@ -522,6 +529,59 @@ def action_view_partner_invoices(self):
...
+
+        return vals
+
```
ecino: The return is not needed, the dictionary is modified in place.

ecino (changes requested):

partner_compassion/models/partner_compassion.py
auclerc: To use the .is_valid_number(), we have to use a parsed number.

partner_compassion/models/partner_compassion.py
auclerc: Yes, correct

partner_compassion/models/partner_compassion.py
```diff
@@ -522,6 +529,66 @@ def action_view_partner_invoices(self):
...
+            61,
+            62,
+            71,
+        ]
+
+        # Destination codes are the first two digits of a mobile phone number
+        all_swiss_mobile_destination_codes = [74, 75, 76, 77, 78, 79]
+
+        # Check if the partner country is Switzerland
+        if vals.get('country_id') == 43 or self.country_id == 43:
```
ecino:
```suggestion
        swiss_country = self.env.ref("base.ch")
        if vals.get('country_id') == swiss_country.id or self.mapped("country_id") == swiss_country:
```

partner_compassion/models/partner_compassion.py
```diff
@@ -522,6 +529,66 @@ def action_view_partner_invoices(self):
...
+
+            if phone:
+                parsed_phone = phonenumbers.parse(phone, "CH")
+                if not phonenumbers.is_valid_number(parsed_phone):
+                    raise UserError(_("Phone number is not valid."))
+                phone_national_destination_code = int(str(parsed_phone.national_number)[:2])
+                if phone_national_destination_code in all_swiss_mobile_destination_codes:
+                    vals["mobile"] = phone
+                    phone_moved_to_mobile = True
+                    # if not mobile:
```
ecino:
```suggestion
```

ecino (changes requested):

ecino:
Thank you, I merged manually (for fixing conflicts).

PR #1592, opened 2024-04-02
change bvr ref in the payment ref and add readable customer ref to be…

davidwul:
… used en direct debit/lsv

TODO: enhance or rewrite the textual ref produced @NoeBerdoz  this might be a zivi job

This change enable not to port the lsv_compassion module


NoeBerdoz:
After talking with David, he will create a task for this case, the assignee will create a new pull request based on this one to remain consistent in the code history.

langjer:
Done in #1593 PR can be close

PR #1606, opened 2024-05-27
fixed error in atachement of disaster alert

minedcp:
added suod() in function _render_qweb_pdf.


ecino:
What was the access error message ? We might miss a security access instead.

minedcp:
image
this was the message error

minedcp:
image
here is the error message.

minedcp:
corrected the code by removing the sudo

ecino (approved):

PR #1607, opened 2024-05-27
added sudo().contact_create() & invisible duplicate tags

minedcp:
added the sudo() function to iap to be able to write modification and put the tags invisible for duplicates. only visible now from button Check duplicates.

### DIFF
--- a/sms_939/models/iap_account.py
+++ b/sms_939/models/iap_account.py
@@ -40,3 +40,7 @@ class IapAccount(models.Model):
             }
         )
         return res
+
+    def write(self, vals):
+        # Ensure sudo() is used for elevated permissions
+        return super(IapAccount, self.sudo()).write(vals)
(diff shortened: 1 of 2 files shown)


sms_939/models/iap_account.py
```diff
@@ -40,3 +40,7 @@ def _server_env_fields(self):
             }
         )
         return res
+
+    def write(self, vals):
+        # Ensure sudo() is used for elevated permissions
+        return super(IapAccount, self.sudo()).write(vals)
```
ecino, line 46: In that case it's better to add security access to all users instead of doing a sudo here. It's also strange that the SMS accounts needs to be written. Maybe you should debug why the write method is called and maybe fix that instead.

minedcp: i will find out why the write method is being called then thank you.

minedcp: changed the name of the branch to the correct task FIX-T1294-Creating-new-contact

PR #1612, opened 2024-06-03
FIX send communication SMS

davidwul:
T1312


ecino (commented):
You don't call anymore the super, are we sure it's ok? Should we still call it at the end?

davidwul:
@ecino there is no super() this function may have disapeared with the migration

PR #1623, opened 2024-06-20
missing onchange decorator

davidwul:
decorator missing to pay the june salary :)


ecino (approved):
If my salary depends on it, I will merge this asap.

PR #1624, opened 2024-06-20
14 payroll decorator

davidwul:
use a compute field insead of onchange


hr_switzerland/models/hr_payslip.py
```diff
@@ -37,10 +37,11 @@ def action_payslip_done(self):
...
         for payslip in self:
-            if payslip.pay_13_salary:
-                payslip.amount_13_salary = payslip.contract_id.provision_13_salary
-            else:
-                payslip.amount_13_salary = 0
+            if payslip.state == 'draft':
+                if payslip.pay_13_salary:
+                    payslip.amount_13_salary = payslip.contract_id.provision_13_salary
+                else:
+                    payslip.amount_13_salary = 0
```
ecino, line 48: Compute methods should assign a value for every case otherwise it may crash.
```suggestion
            if payslip.state == 'draft':
                if payslip.pay_13_salary:
                    payslip.amount_13_salary = payslip.contract_id.provision_13_salary
                else:
                    payslip.amount_13_salary = 0
            else:
                payslip.amount_13_salary = payslip.amount_13_salary
```

ecino (changes requested):

hr_switzerland/models/hr_payslip.py
davidwul: @ecino 
if the payslip is not in draft, we don't want to change the amount_13_salary, if there was someting we don't want it to be changed

davidwul: could we assign the same value:
`payslip.amount_13_salary = payslip.amount_13_salary`

ecino: No, because you change to compute method, all values will be erased, except if you add the `store=True` attribute

hr_switzerland/models/hr_payslip.py
```diff
@@ -20,7 +20,7 @@ class HrPayslip(models.Model):
     )
 
     amount_13_salary = fields.Float(
-        string="13th salary to add", digits=dp.get_precision("Account")
+        string="13th salary to add", digits=dp.get_precision("Account"),compute="_compute_13_salary"
```
ecino:
```suggestion
        string="13th salary to add", digits=dp.get_precision("Account"),
        compute="_compute_13_salary", store=True
```

hr_switzerland/models/hr_payslip.py
```diff
@@ -37,10 +37,11 @@ def action_payslip_done(self):
                 move.action_post()
         return res
 
-    @api.onchange('employee_id', 'pay_13_salary', 'contract_id')
+    @api.depends('employee_id', 'pay_13_salary', 'contract_id')
```
ecino:
```suggestion
    @api.depends('employee_id', 'pay_13_salary', 'contract_id', 'state')
```

hr_switzerland/models/hr_payslip.py
ecino: I updated my suggestion.

ecino (changes requested):

ecino (approved):

PR #1629, opened 2024-07-09
[T1506] Add employee image as link in email's signature 

NoeBerdoz:
Related ticket: [T1506](https://erp.compassion.ch/web#id=2098&action=521&active_id=895&model=project.task&view_type=form&cids=1&menu_id=2029)

This PR changes how the employee signature image is handled when sending an email.
Instead of using a base64 encoded image to display the employee image, it retrieves the it from a new created endpoint 
**/employee/image/:employee_id.**

This change is needed because integrated base64 encoded image are not supported by many mail clients, as for example Gmail.

### DIFF
--- a/partner_communication_switzerland/controllers/employee_image.py
+++ b/partner_communication_switzerland/controllers/employee_image.py
@@ -7,46 +7,41 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
+import base64
+
+from werkzeug.exceptions import NotFound
+
 from odoo import http
 from odoo.http import request
-from werkzeug.exceptions import NotFound
-import base64
 
 
 class EmployeeImageController(http.Controller):
-    @http.route(
-        '/employee/image/<int:employee_id>/',
-        auth='public',
-        type='http',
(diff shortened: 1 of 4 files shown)


partner_communication_switzerland/controllers/employee_image.py
```diff
@@ -0,0 +1,54 @@
...
+from odoo.http import request
+
+
+class EmployeeImageController(http.Controller):
+    @http.route(
+        "/employee/image/<int:employee_id>/",
+        auth="public",
+        type="http",
+        methods=["GET"],
+        website=True,
```
ecino: I don't think we need the website parameter if it's a URL not used on the website

partner_communication_switzerland/controllers/employee_image.py
```diff
@@ -0,0 +1,54 @@
...
+        sitemap=False,
```
ecino: Therefore this won't be needed if the website parameter is removed

ecino (changes requested):

PR #1633, opened 2024-07-18
T1634 - Fix faulty email templates

Prazn:
# Description
It seems that some email templates do not work anymore. This is due to the migration to odoo 14, the model `survey.survey` which is used in some templates [does not have the field ](https://github.com/odoo/odoo/commit/84bb9c748db26656d3ecd062183afe5f042dc569#diff-3394f3b291a5df9e6e82724c34b8e14b6835bc84c98fa75a5400d5237d4a3005) `public_url` anymore.

# Technical Aspects
The fix is pretty straightforward, the field `public_url` was replaced by a method `get_start_url()`.
So we added a new migration to update the templates in the `mail_template` table. We also needed to fix the corresponding translations of those templates in the `ir_translation` table since they also contain the `public_url`.


ecino (commented):
One simple question: are we sure all templates having `public_url` refer to surveys and should be replaced? I just want to make sure we won't replace other usages that could still be valid.

Prazn:
> One simple question: are we sure all templates having `public_url` refer to surveys and should be replaced? I just want to make sure we won't replace other usages that could still be valid.

Unfortunately I could not find a way to be more specific in the query, since sometimes we wanted to replace `object.env.ref('muskathlon.muskathlon_form').with_context(relative_url=True).public_url` and other times `object.env.ref('partner_communication_switzerland.wrpr_onboarding_survey').public_url`.
I think there were only 5 email templates with `public_url` in the body, and when I looked at them they did need to be replaced by `get_active_url()` since they did not work anymore.

However in the second query to fix the translations, I could be specific and change only those that were modified by the previous query to not have unwanted side effects

PR #1638, opened 2024-07-31
T1486 can not log in 2fa activated

clementcharmillot:
This is a draft to support token-based authentication on the Translation Platform.

Currently self-contained in its own module for simplicity purpose.

# TODO before merging

- [x] Write unit / integration tests to verify the security of the implementation
- [x] Fix points marked "TODO" in the code
- [x] Improve documentation

## Related PR to modify the frontend
CompassionCH/translation-platform-web#41
Note: **The current PR must be merged/deployed before the related frontend PR**. Otherwise, the frontend will make requests that the backend doesn't understand.


auth_external/controllers/auth.py
```diff
@@ -0,0 +1,69 @@
...
+    def login(self):
+        username = request.jsonrequest["username"]
+        password = request.jsonrequest["password"]
+        totp = request.jsonrequest["totp"]
+
+        db = request.env.cr.dbname
+        res_users = registry(db)["res.users"]
+
+        user_id = res_users.authenticate(
+            db, username, password, {"totp": totp, "interactive": False}
```
nlachat-compassion, line 28: To test: will this work? Not sure if the auth_totp module is designed for this.

nlachat-compassion:
I tested this PR locally. 
Authentication for users with 2FA enabled seems to work with the modified frontend ( https://github.com/CompassionCH/translation-platform-web/pull/41 ). 
Looking at the code, the implementation seems secure. However, I would like to write unit/integration tests to increase certainty, as this is very sensitive code.

nlachat-compassion:
In the process of trying to store the tokens in HttpOnly cookies, I discovered that the XmlRpcClient library used by the frontend (`import { XmlRpcClient } from '@foxglove/xmlrpc';`) does not support the `withCredentials` property (https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials). This means that the access token is not being sent with XmlRpc requests, and thus the server rejects them with AccessDenied. It would require substantial efforts to use another library or to re-implement the library with the required withCredentials property. 
For now, I will revert the changes made in order to use cookies for the tokens. Maybe in the future, the transition to a new XmlRpc library can be made, and thus the tokens could be stored in cookies, which is more secure than the current implementation using localStorage.

nlachat-compassion:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=CompassionCH_compassion-switzerland&pullRequest=1638) **Quality Gate failed**
> 
> Failed conditions  [5 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=CompassionCH_compassion-switzerland&pullRequest=1638&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=CompassionCH_compassion-switzerland&pullRequest=1638)

These are false positives.

nlachat-compassion:
Pre-commit is failing because of setup files:
```
setup/account_reconcile_checkout/odoo/addons/account_reconcile_checkout
setup/account_reconcile_checkout/setup.py
setup/auth_external/odoo/addons/auth_external
setup/auth_external/setup.py
setup/crm_switzerland/odoo/addons/crm_switzerland
setup/crm_switzerland/setup.py
```
I don't know what they're used for. Should they be committed?

NoeBerdoz (approved):

PR #1646, opened 2024-08-16
T1712 - Delay automatic sub activation (1/2)

Prazn:
# Description
The activation of a sub sponsorship was too fast for some sponsors, it did not allow then enough time to delete the permanent order if they wanted to.
What was decided is that for a sub sponsorship in waiting, it is activated after the second payment. So if two invoices have been paid for, we consider that the sponsor wants to sponsor the new child.

# Technical Aspects
We changed the callback method `invoice_paid` of `RecurringContract` so that the condition to activate the contract after an invoice has been paid can be overridden (see https://github.com/CompassionCH/compassion-accounting/pull/241). 

Before it would always activate the contract if it was in `waiting`. And now in this case we override that condition to also check that there are at least two invoices after the sub proposal has been sent which have been paid for.


partner_communication_switzerland/models/contracts.py
```diff
@@ -548,3 +548,21 @@ def _send_new_dossier(self, correspondent=True):
...
+                    ("invoice_date_due", ">=", sub_proposal_date),
+                    ("payment_state", "=", "paid"),
+                    ("partner_id", "=", contract.partner_id.id),
+                    ("line_ids.contract_id", "=", contract.id),
+                ]
+            )
+
+            return len(paid_invoices) >= 2
+
+        return super()._can_activate_contract(contract)
```
```suggestion
    def _activable_contracts(self):
        """Activate sub proposal contract if it is waiting for payment and
        at least two invoices have been paid."""
        is_ok = super()._activable_contracts()
        for sub in self.filtered("sub_proposal_date"):
            sub_proposal_date = contract.sub_proposal_date.replace(day=1)
            paid_invoices = self.env["account.move"].search_count(
                [
                    ("invoice_date_due", ">=", sub_proposal_date),
                    ("payment_state", "=", "paid"),
                    ("partner_id", "=", contract.partner_id.id),
                    ("line_ids.contract_id", "=", contract.id),
                ]
            )
            if paid_invoices < 2:
                is_ok -= sub
        return is_ok
```

ecino (changes requested):
Adapt for proposed changes in parent module.

partner_communication_switzerland/models/contracts.py
Prazn: Applied your suggestion (I just adapted it a bit so that we iterate on activatable contracts from the `super()` call)

ecino (approved):

PR #1647, opened 2024-08-19
T1702 - Donations communication test cases

clementcharmillot:
As the test partner is picked randomly, it might not have any line we can use for displaying. Simply don't restrict the test lines from belonging to the test partner.

### DIFF
--- a/partner_communication_switzerland/models/partner_communication_config.py
+++ b/partner_communication_switzerland/models/partner_communication_config.py
@@ -7,9 +7,13 @@
 #    The licence is in the file __manifest__.py
 #
 ##############################################################################
+import logging
 import random
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+_logger = logging.getLogger(__name__)
 
 
 class PartnerCommunication(models.Model):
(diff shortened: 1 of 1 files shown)


ecino:
Wouldn't it be better to apply a domain to the partner field so that the user can only select partners that have move lines ? Because with your solution, we will select a partner for testing and then the generated communication will be for another one, which will be confusing.

clementcharmillot:
I've implemented your suggestion for the default random selection of a partner. 
But for the partner input, instead of filtering partners, I display an error if the manually picked partner has no invoice of type fund. As a partner having such invoice is not obvious, it might be confusing if it simply doesn't appear in search without explanation when we are searching for a specific one.
Let me know if you'd prefer the other approach.

partner_communication_switzerland/models/partner_communication_config.py
```diff
@@ -141,6 +142,13 @@ def _get_test_objects(self, partner, children=None):
                 )
                 .ids
             )
+            if len(object_ids) == 0:
+                raise UserError(
+                    _(
+                        "The selected partner has no invoice line but this test case "
+                        "requires one. Please, pick a partner with an invoice line."
```
ecino:
```suggestion
                        "The selected partner is not a donor but this test case "
                        "requires one. Please, pick a donor."
```

ecino (changes requested):
OK, but the error message will be more explicit to our staff like that.

PR #1656, opened 2024-09-06
T1799 - Fix sponsorship onboarding form

Prazn:
# Description
New sponsors are invited to a visio conference through a communication with a link to a form where they can register, which will add a new `partner_zoom_attendee`. However if the participant is already invited and then clicks on the form an sql exception will be raised because the participant already exists.

# Technical Aspects
Two fixes have been made : 
* The registration form will now set the parcipant in the state `confirmed` if they click on the "Register" button
* When we create a new `partner_zoom_attendee`, if he already exists we simply update the information

### DIFF
--- a/partner_communication_switzerland/models/res_partner_zoom_attendee.py
+++ b/partner_communication_switzerland/models/res_partner_zoom_attendee.py
@@ -86,12 +86,18 @@ class ZoomAttendee(models.Model):
 
         for vals in vals_list:
             existing_attendee = self.search(
-                [('partner_id', '=', vals.get('partner_id')),
-                 ('zoom_session_id', '=', vals.get('zoom_session_id'))])
+                [
+                    ("partner_id", "=", vals.get("partner_id")),
+                    ("zoom_session_id", "=", vals.get("zoom_session_id")),
+                ]
+            )
             if existing_attendee:
                 vals_list_to_create.remove(vals)
-                filtered_vals = {k: v for k, v in vals.items() 
-                                 if k not in ['partner_id', 'zoom_session_id']}
+                filtered_vals = {
+                    k: v
+                    for k, v in vals.items()
+                    if k not in ["partner_id", "zoom_session_id"]
+                }
                 existing_attendee.update(filtered_vals)
                 res.append(existing_attendee)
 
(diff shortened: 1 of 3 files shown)


partner_communication_switzerland/models/res_partner_zoom_attendee.py
```diff
@@ -81,7 +81,28 @@ def _expand_states(self, states, domain, order):
 
     @api.model_create_multi
     def create(self, vals_list):
-        res = super().create(vals_list)
+        res = []
```
ecino:
```suggestion
        res = self
```

partner_communication_switzerland/models/res_partner_zoom_attendee.py
```diff
@@ -81,7 +81,28 @@ def _expand_states(self, states, domain, order):
...
+                ]
+            )
+            if existing_attendee:
+                vals_list_to_create.remove(vals)
+                filtered_vals = {
+                    k: v
+                    for k, v in vals.items()
+                    if k not in ["partner_id", "zoom_session_id"]
+                }
+                existing_attendee.update(filtered_vals)
```
ecino:
```suggestion
                del vals["partner_id"]
                del vals["zoom_session"]
                existing_attendee.write(vals)
```

partner_communication_switzerland/models/res_partner_zoom_attendee.py
```diff
@@ -81,7 +81,28 @@ def _expand_states(self, states, domain, order):
...
+                res.append(existing_attendee)
```
ecino:
```suggestion
                res += existing_attendee
```

partner_communication_switzerland/models/res_partner_zoom_attendee.py
```diff
@@ -81,7 +81,28 @@ def _expand_states(self, states, domain, order):
...
+
+        res.append(super().create(vals_list_to_create))
```
ecino:
```suggestion
        res += super().create(vals_list_to_create)
```

ecino (changes requested):
You also need to whitelist the field `state` for web forms. This is done in file `/data/form_data.xml`

partner_communication_switzerland/models/res_partner_zoom_attendee.py
Prazn: Done

partner_communication_switzerland/models/res_partner_zoom_attendee.py
Prazn: Done

ecino (approved):

PR #1659, opened 2024-09-11
T1805 - Fix signup email signature

Prazn:
# Description
When a user creates a new account, he receives an email to confirm his registration, however the email was signed by the user himself.

# Technical Aspects
It look like for the email signup template (as well as some others), when we generate the email the `self.env.user` is used to sign the email. This is a bit of a side effect of the fact that the `user` passed is defined in `mail_render_mixin._render_jinja_eval_context()` of the `mail` module, but the signature is something that we added later in `partner_communication_switzerland`. 

To fix this we have a bit of the workaround,  but the idea is that when we generate the signature we first check if the user passed is an employee, if it is not the case, we generate the signature of the user's company (which is always Compassion).


partner_communication_switzerland/models/res_users.py
```diff
@@ -64,38 +64,62 @@ def _compute_signature(self):
...
-                    "phone_link": phone_link.get(lang),
-                    "phone": phone.get(lang),
-                    "mobile": employee.mobile_phone,
-                    "mobile_link": (employee.mobile_phone or "")
-                    .replace(" ", "")
-                    .replace("(0)", ""),
-                    "facebook": facebook.get(lang),
-                    "employee_image_url": employee_image_url,
-                }
+                if not employee:
```
ecino:
```suggestion
                if employee:
```

ecino (changes requested):

partner_communication_switzerland/models/res_users.py
Prazn: oops forgot to revert this condition, was used for testing

ecino (approved):

PR #1663, opened 2024-09-19
T0024 - Automatic reminder for invalid partners

Gordack:
A scheduled action should be created, similar to : 
image

After test with MailHog, the following output is obtained : 
image


partner_communication_switzerland/data/auto_reminder_archive_contact.xml
```diff
@@ -0,0 +1,30 @@
...
+            <field
+        name="subject"
+      >Reminder: Invalid partners</field>
+            <field name="body_html" type="html">
+                <div>
+                    <b>The following partners meet the criteria and are considered invalid : </b>
+                    % set partner_info = object.partner_id.get_archive_contact()
+                    <span style="font-family: 'miller';">
+                    % for partner in partner_info:
+                        <p>ID ${partner.id} - ${partner.name}</p>
```
ecino: I think that the partner reference is more meaningful to SDS than the database ID. You could also include a direct link to the contact:
```suggestion
                        <p><a href="/web#id=${partner.id}&model=res.partner&view_type=form&menu_id=1744">Ref ${partner.ref} - ${partner.name}</a></p>
```

ecino (changes requested):
One small remark and another general question : Could we simply automatically archive the found contacts (and still send the notification email to SDS)? Then in the email we could simply write: "Those following contacts were found and automatically archived". That means writing {"archive": True} on them.

Gordack:
> One small remark and another general question : Could we simply automatically archive the found contacts (and still send the notification email to SDS)? Then in the email we could simply write: "Those following contacts were found and automatically archived". That means writing {"archive": True} on them.

When I discussed that with Sarah, she was not so happy about the idea of automating the archive part, because she was scared that Odoo would archive people that should not be archived. Therefore, when we discussed that with Andre, we choose to not automatically archive first, in order to let SDS check that the people we propose to archive are the good ones. But yes, if SDS is happy with how the criteria are set, it would be better to archive the people that need to be archived automatically.

ecino:
Waiting Sarah's feedback in order to know whether we should directly archive those contacts or not.

nlachat-compassion:
> Waiting Sarah's feedback in order to know whether we should directly archive those contacts or not.

@ecino I'm taking over this task. So in the end, we should **not** automatically archive the invalid partners, correct? Only send an email reminder with the list of potentially invalid partners?

ecino:
@nlachat-compassion exactly

nlachat-compassion:
What I did until now:
- Add xml definition for the cron job which sends the reminder (configured to send once a week)
- Add a link to the task T0024 as a reference for the criteria which define a partner as invalid in the email reminder
- Pass the list of invalid partners through the context in the email template in order to avoid computing it twice. This requires a modification to another repo: https://github.com/CompassionCH/compassion-modules/pull/1979 , which should be merged **before** this one.

@ecino what do you think about my proposed implementation?

ecino:
@nlachat-compassion This seems ok for me. Did you test without the modification in compassion-modules? I have the feeling the context should be passed even without the other change.

nlachat-compassion:
> @nlachat-compassion This seems ok for me. Did you test without the modification in compassion-modules? I have the feeling the context should be passed even without the other change.

Ah yes, indeed @ecino it works without the change to `compassion-modules`. I think I probably forgot to update the view at some point and I thought that the context was not passed properly.

ecino (approved):

PR #1674, opened 2024-11-22
With the new strict ref matching, overriding the partner_ref with the…

davidwul:
… product name hinder the reconciliation.


davidwul:
j'ai déjà mis la modif en prod

ecino (approved):

PR #1679, opened 2025-01-24
[T2128] FEAT: Event registration, down payment invoice sent with confirmation.

auclerc:
The task has another PR in compassion-website: https://github.com/CompassionCH/compassion-website/pull/60

### DIFF
--- a/website_switzerland/models/event_registration.py
+++ b/website_switzerland/models/event_registration.py
@@ -6,12 +6,10 @@ class EventRegistration(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        vals_list[0]["stage_id"] = self.env.ref("website_switzerland.stage_group_pay").id
-
         registrations = super().create(vals_list)
         activate_task = self.env.ref("website_switzerland.task_activate_account")
         child_protection_task = self.env.ref(
-           "website_switzerland.task_sign_child_protection"
+            "website_switzerland.task_sign_child_protection"
         )
         for registration in registrations:
             partner = registration.partner_id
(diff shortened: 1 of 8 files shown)


website_switzerland/models/event_registration.py
```diff
@@ -6,10 +6,12 @@ class EventRegistration(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
+        vals_list[0]["stage_id"] = self.env.ref("website_switzerland.stage_group_pay").id
+
```
ecino:
```suggestion
```

website_switzerland/models/event_registration.py
```diff
@@ -21,6 +23,12 @@ def create(self, vals_list):
                 registration.task_ids.filtered(
                     lambda t, m_task=child_protection_task: t.task_id == m_task
                 ).write({"done": True})
+
+            if not partner.country_id:
+                partner.country_id = 44
+
```
ecino, line 27: I would prefer that we do this only for the specific event type. Other event types might have a different workflow. But the best would be a general behaviour that works for every event type. The method [_default_stage](https://github.com/compassionch/compassion-website/blob/51aac75c564499e9d68e14e7bb7a23f22c61a036/website_event_compassion/models/event_registration.py#L327) have this in mind but doesnt work for frontend submitted registrations, because the context is not set.

website_switzerland/models/event_registration.py
```diff
@@ -21,6 +23,12 @@ def create(self, vals_list):
                 registration.task_ids.filtered(
                     lambda t, m_task=child_protection_task: t.task_id == m_task
                 ).write({"done": True})
+
+            if not partner.country_id:
+                partner.country_id = 44
```
ecino:
```suggestion
                partner.country_id = self.env.ref("base.ch")
```

ecino (changes requested):

PR #1680, opened 2025-02-03
[T2142] FEAT: Add field Title for new sponsors on donations

auclerc:


ecino (approved):
I just changed the code for dynamically injecting existing titles instead of hardcoding them. It lets us adapt the titles we want to display and can also work on other databases if needed.

PR #1688, opened 2025-05-06
[T2087] bvr layout fixed for all languages

Nath-S07:
I checked the BVR layout for the other languages, not only for german. So depending on the partner's sponsorship for any language it would get out of the bill. For that, I adjusted the margin for the additional information section. This ensures that the text wraps correctly and never extends outside the printable area, regardless of the language or text length.


report_compassion/report/bvr_layout.xml
```diff
@@ -196,6 +196,8 @@
                             <span
               class="swissqr_text title"
             >Additional information</span><br />
+
+            <div style="max-width: 70mm; overflow-wrap: break-word;">
```
NoeBerdoz: I think the break-word will decrease the read smoothness, let's remove it.
```suggestion
            <div style="max-width: 70mm;">
```

NoeBerdoz (changes requested):
The branch has a minor conflict, please fix it

NoeBerdoz (approved):

PR #1689, opened 2025-06-10
[T2281]  remove missleading footer

SamBachmann:
I adapted the footer to be disabled when the connected user is already a sponsor.
To ensure proper reloading when we update the module, I also set active attribute of template to true.


ecino (approved):
@NoeBerdoz Code looks good to me, but should we have a special branch for MC2 here as well ? Or can we put this change in production before MC2?

PR #1692, opened 2025-07-22
T2515 feat: select neediest csp country when country null

NoeBerdoz:
### Feat: Automatic CSP Country Selection Based on Needs

**Ticket: [T2515](https://erp.compassion.ch/web#id=3107&action=521&active_id=895&model=project.task&view_type=form&cids=1&menu_id=2029)**

---

#### Summary

This pull request introduces a new intelligent logic to automatically assign a CSP sponsorship to the country with the most need when one is not specified in the incoming email.

Previously, if an email was received without a designated country, the process would fail. This change makes the system more robust and ensures that new CSP are routed to the areas where they will have the most impact.

---

#### Changes

* **New Method: `_get_neediest_csp_country_code()`**
    * A new method has been created to determine the best country for a new CSP sponsorship.
    * The selection is based on a two-factor sorting key:
        1.  It prioritizes countries with the **lowest percentage of used slots**.
        2.  It selects the country with the **highest absolute number of available slots**.

* **Continent Support in `_parse_csp_info()` and fix ReDoS vulnerability**
    * The email parsing logic has been enhanced to recognize and map the "Continent" field from the email body across multiple languages (FR, DE, IT).
    * Hardened all regular expressions by replacing the vulnerable `(.*?)` pattern with a non-backtracking `([^\n]*)` pattern. This resolves a ReDoS (Regular Expression Denial of Service) vulnerability and improves parsing performance.

* **Fix in `message_new()` for `recurring.contract`**
    * The core logic has been updated to use the new features.
    * If an incoming email for a CSP sponsorship does **not** specify a country, the system now calls `_get_neediest_csp_country_code()` to automatically and intelligently assign one.

### DIFF
--- a/wordpress_connector/models/contracts.py
+++ b/wordpress_connector/models/contracts.py
@@ -32,55 +32,53 @@ SPONSOR_MAPPING = {
 }
 
 CONTINENT_MAPPING = {
-            # French
-            'Asie': 'Asia',
-            'Afrique': 'Africa',
-            'Amérique Latine & Caraïbes': 'LATAM',
-
-            # German
-            'Asien': 'Asia',
-            'Afrika': 'Africa',
-            'Lateinamerika & Karibik': 'LATAM',
-
-            # Italian
-            'Asia': 'Asia',
-            'Africa': 'Africa',
-            'America Latina e Caraibi': 'LATAM',
-        }
+    # French
+    "Asie": "Asia",
+    "Afrique": "Africa",
+    "Amérique Latine & Caraïbes": "LATAM",
+    # German
+    "Asien": "Asia",
+    "Afrika": "Africa",
+    "Lateinamerika & Karibik": "LATAM",
+    # Italian
+    "Asia": "Asia",
+    "Africa": "Africa",
+    "America Latina e Caraibi": "LATAM",
+}
 
 CONTINENT_COUNTRY_MAP = {
-    'Africa': [
-        'BF',  # Burkina Faso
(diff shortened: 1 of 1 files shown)


wordpress_connector/models/contracts.py
```diff
@@ -458,7 +518,42 @@ def _parse_csp_info(self, data_string):
...
+        csp_products = self.env["product.product"].search(
+            [("default_code", "in", [f"csp_{code}" for code in country_codes])]
+        )
+
+        # Sort csp_products by slot_used asc and survival_slot_number desc
+        csp_sorted_by_needs = sorted(
+            csp_products, key=lambda c: (c.slot_used, -c.survival_slot_number)
+        )
+
+        return csp_sorted_by_needs[0].default_code.split("_")[-1]
```
ecino: Since `slot_used` and `survival_slot_number` are stored, doing all in the search method will be more efficient.
```suggestion
        csp_sorted_by_needs = self.env["product.product"].search(
            [("default_code", "in", [f"csp_{code}" for code in country_codes])],
            order="slot_used ASC, survival_slot_number DESC",
            limit=1
        )
        return csp_sorted_by_needs.default_code.split("_")[-1]
```

ecino (changes requested):

wordpress_connector/models/contracts.py
NoeBerdoz: Thank you, that's way better

ecino (approved):
Looks good to me!

PR #1732, opened 2026-01-26
[T2925] FIX: images are now not skewed on volunteering page

Danielgergely:
FIX: added object-fit: cover to the images


NoeBerdoz (approved):
Great, thanks!

PR #1736, opened 2026-01-27
[T2939] - Fix the creation of the duplicates Together engagment type on update

SlWa99:
### Description
This PR fixes critical data consistency issues on `advocate.engagement` records identified during the migration process.

---

### 1. Fix "Together" Duplication (ID 24)

**Issue**  
The historical *Together* record (ID 24) was missing an XML ID. As a result, module updates were creating a duplicate record instead of updating the existing one.

**Fix**  
The pre-migration script now ensures that the correct XML ID is assigned to record ID 24.  
A safety check has also been added to clean up any incorrect links to duplicate records created by previous failed migration runs.

---

### 2. Fix "Prayer" Unique Constraint Crash (ID 19)

**Issue**  
The *Prayer* record (ID 19) existed in an inconsistent state, with both a legacy `__export__` XML ID and the official module XML ID assigned.  
This caused a `psycopg2.errors.UniqueViolation` when attempting to rename the export ID during migration.

**Fix**  
The script now detects this specific conflict.  
If the official XML ID is already present on record ID 19, the obsolete `__export__` ID is safely removed instead of triggering a database crash.

---

### DIFF
--- a/my_compassion_switzerland/migrations/14.0.1.0.1/pre-migration_engagement_type.py
+++ b/my_compassion_switzerland/migrations/14.0.1.0.1/pre-migration_engagement_type.py
@@ -44,6 +44,21 @@ def migrate(env, version):
     # -------------------------------------------------------------------------
 
     # Add missing XML_ID for Together engagement type
+    # Check if 'engagement_together' is already assigned to a WRONG ID
+    cr.execute(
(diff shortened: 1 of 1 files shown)


NoeBerdoz (approved):
Thanks!

